PR #2, opened 2021-01-27
[Huachen] iP

huachen24:


src/main/java/Duke.java
```diff
@@ -47,50 +48,52 @@ public static void addTodo(String todo) {
...
+        }
+        catch (ArrayIndexOutOfBoundsException e) {
+            System.out.println(line);
+            System.out.println("\tPlease specify deadline.");
+            System.out.println(line + "\n");
+        }
+    }
 
     public static void done(String item) {
         int i = Integer.parseInt(item)-1;
```
nivikcivik: maybe could be more descriptive? like "taskDone", etc.

nivikcivik: since the "i" here means something, maybe it can be named as "index" or something more meaningful
Generally, I couldn't find much coding violations! It looks very neat:)

src/main/java/Duke.java
```diff
@@ -36,7 +36,8 @@ public static void list() {
         System.out.println(line);
```
nivikcivik: maybe "line" can be replaced with "dottedLine"? I was a bit confused as to what this was referring to.

src/main/java/Duke.java
```diff
@@ -47,50 +48,52 @@ public static void addTodo(String todo) {
...
     	    System.out.println(line + "\n");
-		}
-		catch (ArrayIndexOutOfBoundsException e) {
-			System.out.println(line);
-			System.out.println("\tPlease specify event date.");
-			System.out.println(line + "\n");
-		}
-	}
+        }
+        catch (ArrayIndexOutOfBoundsException e) {
```
nivikcivik: maybe you could bring the catch statement to the } bracket of your try statement ( like "} catch (...) {" ) I think that is a violation that you could change:)

src/main/java/Duke.java
```diff
@@ -47,50 +48,52 @@ public static void addTodo(String todo) {
...
+            tasks[currentTask] = t;
+            currentTask++;
+            System.out.println(line);
+            System.out.println("\tGot it. I've added this task:");
+            System.out.println("\t   [" + t.getType() + "][" + t.getStatusIcon() + "] "
+                    + t.getDescription() + " " + t.getDate());
+            System.out.println("\tNow you have " + Integer.toString(currentTask) + " tasks in the list.");
+            System.out.println(line + "\n");
+        }
+        catch (ArrayIndexOutOfBoundsException e) {
```
nivikcivik: again here you should move the catch statement to the brackets, but otherwise I can't see any issues:)

nivikcivik (commented):
Overall your code is very well written! I almost couldn't find any issues:) just some very minute name changes and bracket changes, that's all:)

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
+import java.util.Scanner;
+
 public class Duke {
+    public static String line = "\t________________________________________";
```
MingShun98: constant terms should be named in all caps, a different name would be also preferrable

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+                addTodo(command.split(" ", 2)[1]);
+            } else if (command.startsWith("event")) {
+                addEvent(command.split(" ", 2)[1]);
+            } else if (command.startsWith("deadline")) {
+                addDeadline(command.split(" ", 2)[1]);
+            }
+        }
+    }
+
+    public static void bye() {
```
MingShun98: a different preferred like printBye would be better since the variable name may be confusing in the above method

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+        System.out.println("\tHere are the tasks in your list:");
+        for (int i=0; i<currentTask; i++) {
+            System.out.println("\t" + Integer.toString(i+1) + ".[" + tasks[i].getType() + "]["
+                    + tasks[i].getStatusIcon() + "] " + tasks[i].getDescription() + " " + tasks[i].getDate());
+        }
+        System.out.println(line + "\n");
+    }
+
+    public static void addTodo(String todo) {
+        Todo t = new Todo(todo);
```
MingShun98: a different variable name could be used instead of t, may lead to confusion further along the code

src/main/java/Duke.java
```diff
@@ -1,10 +1,121 @@
...
+            Event t = new Event(description, date);
+            tasks[currentTask] = t;
+            currentTask++;
+            System.out.println(line);
+            System.out.println("\tGot it. I've added this task:");
+            System.out.println("\t   [" + t.getType() + "][" + t.getStatusIcon() + "] "
+                    + t.getDescription() + " " + t.getDate());
+            System.out.println("\tNow you have " + Integer.toString(currentTask) + " tasks in the list.");
+    	    System.out.println(line + "\n");
+        }
```
MingShun98: the indentation for try and catch should be
try{
} catch {
}

src/main/java/Event.java
```diff
@@ -0,0 +1,18 @@
+public class Event extends Task {
+    protected String date;
+
+    public Event(String description, String date) {
+        super(description);
+        this.date = date;
+    }
+
```
MingShun98: for indentation leave one line in between methods for shorter code

MingShun98 (commented):
Overall code is neat, some indentation and variable name violation but still a goob job on the code :)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+        case "done":
+            return "done [index of task]";
+        case "delete":
+            return "delete [index of task]";
+        default:
+            return null;
+        }
+    }
+    
+    public static void unknownCommand() {
```
sevenseasofbri: Consider naming the method to something that describes what it does, like `printUnknownCommandResponse()` or something along those lines.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+import java.nio.file.Paths;
+
+public class Duke {
+    public static ArrayList<Task> tasks = new ArrayList<>();
+
+    public static void dividerLine() {
+        System.out.println("\t________________________________________\n");
+    }
+
+    public static void loadData() {
```
sevenseasofbri: This method seems to be a little long, consider extracting and refactoring parts of the code to make it shorter.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+                    if (splitLine[1].equals("1")) {
+                        t.markAsDone();
+                    }
+                } else if (line.startsWith("D")) {
+                    String description = splitLine[2];
+                    String date = splitLine[3];
+                    Deadline t = new Deadline(description, date);
+                    tasks.add(t);
+                    if (splitLine[1].equals("1")) {
+                        t.markAsDone();
```
sevenseasofbri: There seem to be some common statements in each of the blocks - perhaps consider extracting them out to reduce code redundancy? 

https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#minimise-code-duplication

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+                    }
+                }
+            }
+            fileReader.close();
+        } catch (FileNotFoundException e) {
+            ;
+        } catch (IOException e) {
+            System.out.println("\tAn error occured while trying to create a data directory :(");
+            //System.out.println(e.getMessage());
```
sevenseasofbri: Avoid keeping in commented out code, it may confuse you/the reader later on. Consider removing this line.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+            fileWriter.close();
+            System.out.println("\tData saved");
+            return true;
+        } catch (IOException e) {
+            System.out.println("\tAn error occurred while saving your data :(");
+            return false;
+        }
+    }
+
+    public static void takeInput() {
```
sevenseasofbri: Consider naming this to something more intuitive like `readInput` ?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+                dividerLine();
+            } else if (command.equals("save")) {
+                dividerLine();
+                saveData();
+                dividerLine();
+            } else {
+                dividerLine();
+                unknownCommand();
+                dividerLine();
+            }
```
sevenseasofbri: Consider extracting out dividerLine() to the beginning and the end of the while loop. This will reduce redundancy and the number of lines in this function.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+            return null;
+        }
+    }
+    
+    public static void unknownCommand() {
+        System.out.println("\tSorry, this command is not available yet. Stay tuned! :)");
+    }
+    
+    public static void list() {
+        if (tasks.size()==0) {
```
sevenseasofbri: Consider making the happy path prominent in this code by checking whether there are any tasks in the list and returning to the call stack if there aren't any. Then you may eliminate the need for the else case clause and reduce a level of indentation. (There are multiple instances of this in the code so you similarly do this for those instances too)

https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+        } else {
+            try {
+                String description = splitCommand[1].split("/at")[0].strip();
+                String date = splitCommand[1].split("/at")[1].strip();
+                Event t = new Event(description, date);
+                tasks.add(t);
+                System.out.println("\tGot it. I've added this task:");
+                System.out.println("\t   [" + t.getType() + "][" + t.getStatusIcon() + "] "
+                        + t.getDescription() + " " + t.printDate());
+                System.out.println("\tNow you have " + Integer.toString(tasks.size()) + " tasks in the list.");
```
sevenseasofbri: Perhaps use SLAP and extract these statements out into methods? (there are a couple of instances of this in the code, so you can do the similar thing for all of those instances too)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+        } else {
+            return false;
+        }
+    }
+
+    public static void main(String[] args) {
+        dividerLine();
+        System.out.println("\tHello! I'm Duke");
+        System.out.println("\tWhat can I do for you?");
+        dividerLine();
```
sevenseasofbri: Consider extracting this into a method to maintain SLAP.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+            System.out.println("\tAn error occured while trying to create a data directory :(");
+            //System.out.println(e.getMessage());
+        }
+    }
+
+    public static boolean saveData() {
+        try {
+            System.out.println("\tSaving your data...");
+            FileWriter fileWriter = new FileWriter("data/duke.txt");
+            for (int i=0; i<tasks.size(); i++) {
```
sevenseasofbri: Perhaps keep consistent spacing between operands and operations, like `i = 0` instead of `i=0`?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
...
+                dividerLine();
+            } else {
+                dividerLine();
+                unknownCommand();
+                dividerLine();
+            }
+        }
+    }
+    
+    public static String printUsage(String type) {
```
sevenseasofbri: This function doesn't seem to be printing the usage of the command type, so consider renaming it to something like `getUsage`?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,298 @@
+package duke;
+
+import duke.tasks.*;
```
sevenseasofbri, line 3: Avoid using wildcard imports.

sevenseasofbri (commented):
Overall, the readability can be improved by using SLAP and other principles - otherwise well done with following coding standards. 

https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#slap-hard

PR #13, opened 2021-01-29
[Seh Xin Ru] iP

xseh:


src/main/java/Duke.java
```diff
@@ -1,10 +1,342 @@
...
+        return tasks[index];
+    }
+
+    /**
+     * Gets task number from the input string. 
+     * 
+     * @return Task number. 
+     */
+    private static int getTaskNumber(String command) {
+        int taskNumber = Integer.parseInt(command.substring(command.indexOf(" ") + 1)) - 1;
```
brandonfoong: There are multiple nested expressions, you can consider breaking it into multiple expressions for better readability. This will also help when handling exceptions from parseInt() in the future

src/main/java/Deadline.java
```diff
@@ -0,0 +1,19 @@
+public class Deadline extends Task {
+    protected String by;
+
+    /**
```
jonahtwl: Nice use of documentation here to explain the functions and parameters passed into the method!

src/main/java/Duke.java
```diff
@@ -1,10 +1,342 @@
...
+    }
+
+    /**
+     * Displays a given message to the user. 
+     * 
+     * @param message Message to be displayed. 
+     */
+    private static void displayToUser(String... message) {
+        System.out.println(String.format("\t%s", MESSAGE_BORDER));
+        for (String m : message) {
```
brandonfoong: Maybe you can consider using 'message' instead of 'm' as the variable name (to make it more descriptive). Also, since the argument (message) is an array, you can rename it as messages so that it is clearer.

src/main/java/Duke.java
```diff
@@ -1,10 +1,342 @@
...
+            tasks[sizeOfTaskList] = new Deadline(userCommand, date);
+            sizeOfTaskList++;
+            displayAddTaskSuccessMessage();
+        } else {
+            displayToUser(ERROR_INVALID_COMMAND_RECEIVED);
+        }
+    }
+
+    /**
+     * Adds a new event task to task list.
```
brandonfoong: Would be good to keep the length of the comments consistent (e.g. executeAddDeadline and executeAddEvent) since they do similar actions.

src/main/java/Duke.java
```diff
@@ -1,10 +1,342 @@
+import java.util.Scanner;
+
 public class Duke {
+    /**
```
jonahtwl: I am not too sure on this myself but, maybe you can start a comment with the "//"? As "/**" is used for Java documentation on methods?

src/main/java/Duke.java
```diff
@@ -1,10 +1,342 @@
...
+     * Words for commands. 
+     */
+    private static final String COMMAND_TODO_WORD = "todo";
```
jonahtwl: Great use of constants! Do always remember the naming convention as what you are already doing! :)

brandonfoong (commented):
Good job! I like the use of constants for the commands and messages, so that it is easier to modify them in the future (if necessary). There are a few standard violations that I've noticed and commented on, please do take not of those.

src/main/java/Duke.java
```diff
@@ -1,10 +1,342 @@
...
+     * Finds the location of the filters and locate the date from the task. 
+     * Removes the task date information from userCommand, keeping only the task name in userCommand. 
+     * 
+     * @param filterString The string to find in userCommand depending on an event (/at) or deadline (/by). 
+     * @return The date extracted from userCommand. 
+     */
+    private static String processParameters(String filterString) {
+        int indexOfDate = userCommand.indexOf(filterString);
+        if (indexOfDate > 0) {
+            String date = userCommand.substring(indexOfDate + 4);
```
jonahtwl: You may want to use a constant to replace the number 4 here (or maybe a comment to explain) as other's may not understand why we have to offset 4 from the date index itself.

src/main/java/Duke.java
xseh: Thank you, perhaps I could use a comment instead? the main part of getTaskNumber function is to get the task number from the string.

src/main/java/Duke.java
xseh: My other variable name is already "message" so I can't use the same variable name. I am not sure if "String..." is an array or a regular string.

src/main/java/Duke.java
xseh: Oh thanks for noticing, Ill change it so it will look similar

src/main/java/Deadline.java
xseh: Thank you!

src/main/java/Duke.java
xseh: I can clarify that next time, for now I will just change to // as comments. Thank you. 

Edit: I checked against contacts1.java from the exercise which showed that it is okay but I will still change before asking.

src/main/java/Duke.java
xseh: Thank you! I will

src/main/java/Duke.java
xseh: That's true. I will leave a comment indicating the magic number. Thanks.

src/main/java/duke/Constants.java
```diff
@@ -0,0 +1,66 @@
...
+    public static final String ERROR_INVALID_SYNTAX_RECEIVED = "Invalid syntax! Usage: ";
+    public static final String ERROR_IO = "IO Error, please try again. ";
+    public static final String ERROR_IMPORT_TASK = "Failed to import some tasks. ";
+    public static final String ERROR_FILE_NOT_FOUND = "File not found, loading new tasklist.";
+    public static final String ERROR_NO_MATCH_FOUND = "No matches found. Please refine your search";
+    public static final String ERROR_INVALID_DATE_RECEIVED = "Invalid date received. Format: yyyy-mm-dd";
+    public static final String ERROR_INVALID_NUMBER = "Invalid task number received.";
+
+    // Syntax messages for the commands.
+    public static final String MESSAGE_MARK_SYNTAX = " done <task number>";
```
okkhoy, line 41: Good job on writing up the variable names to follow the standard guidelines!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,81 @@
+package duke;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.time.format.DateTimeParseException;
+import duke.error.*;
```
okkhoy: Please import specific classes and not use wildcards. You can get the IDE to prevent importing `*`.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,81 @@
...
+     * @param filePath Location of the storage file. 
+     */
+    public Duke (String filePath) {
+        ui = new Ui();
+        storage = new Storage(filePath);
+        try {
+            tasks = new TaskList(storage.readFromFile());
+        } catch (FileNotFoundException e) {
+            ui.displayToUser(Constants.ERROR_FILE_NOT_FOUND);
+            tasks = new TaskList();
```
okkhoy, line 34: This is good error handling. Keep it up!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,81 @@
...
+        } catch (ImportTaskException exception) {
+            ui.displayToUser(Constants.ERROR_IMPORT_TASK);
+        }
+    }
+
+    /**
+     * Runs the program. 
+     * Gets user input and parses command.
+     * Executes the program according to the command input.
```
okkhoy: Try to be succinct in the header comment.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,181 @@
...
+    protected String[] getCommand(TaskList tasks, String fullCommand) throws NumberFormatException, 
+            InvalidSyntaxException, TaskListEmptyException, IndexOutOfBoundsException, 
+            IllegalCommandException, DateTimeParseException {
+
+        // Split command according to first instance of " "
+        String[] extractedCommands = fullCommand.split(" ", 2);
+
+        // First index (i.e. 0) contains command word, next index contains parameters if any 
+        String command = extractedCommands[Constants.COMMAND_INDEX].toLowerCase().trim();
+        return parseCommand(tasks, command, extractedCommands);
```
okkhoy: This method has things running at multiple levels of abstraction. e.g., splitting to invoking a method. Try to improve SLAP on this one.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,181 @@
...
+    private String[] parseCommand(TaskList tasks, String command, String[] extractedCommands) 
+            throws InvalidSyntaxException, IndexOutOfBoundsException, NumberFormatException, 
+            TaskListEmptyException, IllegalCommandException {
+        switch (command) {
+        case Constants.COMMAND_LIST_WORD:
+        case Constants.COMMAND_EXIT_WORD:
+            // No further parsing needed, return data immediately
+            return setCommandParameters(command, null, null);
+        case Constants.COMMAND_TODO_WORD:
+        case Constants.COMMAND_FIND_WORD:
```
okkhoy: You may be processing the commands similarly, but it is a better idea to invoke the methods separately to avoid misunderstanding by someone who is maintaining the code (which could be you after a few weeks!)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * Gets task number from the input string.
+     * 
+     * @return Task number (Starts with 0).
+     * @throws NumberFormatException If userCommand is not an integer.
+     * @throws IllegalCommandException If no task number is detected.
+     * @throws TaskListEmptyException If task list is empty.
+     */
+    private int getTaskNumber(String command, TaskList tasks, String commandParameter) throws NumberFormatException,
```
okkhoy: The line seems too long.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,181 @@
...
+     * @throws NumberFormatException If userCommand is not an integer.
+     * @throws IllegalCommandException If no task number is detected.
+     * @throws TaskListEmptyException If task list is empty.
+     */
+    private int getTaskNumber(String command, TaskList tasks, String commandParameter) throws NumberFormatException, 
+            TaskListEmptyException {
+        if (tasks.isTaskListEmpty()) {
+            throw new TaskListEmptyException();
+        }
+        // Since user input task number starts from 1, remove 1 from taskNumber to reflect the correct index in tasks.
```
okkhoy: There are two problems.
1. the line is too long.
2. This comment seems to explain to the developer what is happening. Ensure to put it in the javadoc if necessary, or simply use Github comments (like this one) to record such communication. Such comments tend to become obsolete very soon

src/main/java/duke/error/ImportTaskException.java
```diff
@@ -0,0 +1,7 @@
+package duke.error;
+
+/**
+ * Throws if task cannot be imported properly.
```
okkhoy: probably an incorrect javadoc comment. This is the exception that gets thrown, it doesn't throw anything as such

okkhoy (commented):
Please pay attention to slap.
You may want to clean up the code that affect readability
e.g., by fixing misleading comments, and things that may violate line length guideline

PR #19, opened 2021-01-29
[Liu] iP

fsgmhoward:


src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
+        String description = String.join(" ", Arrays.copyOfRange(arguments, 1, arguments.length));
+        tasks.add(new ToDo(description));
+        printNewTask(tasks);
+    }
+
+    // Save a new task with no specific type in the task list
+    protected static void save(Vector<Task> tasks, String description) {
+        tasks.add(new Task(description));
+        printNewTask(tasks);
+    }
```
gcksean: The method name might be a little vague, can consider renaming the method so that it is more clear what it does.

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        // Initialize a vector to store all the tasks
+        Vector<Task> tasks = new Vector<>();
+
+        greeting();
+        printlnWithIndent(LONG_LINE);
+
+        Scanner in = new Scanner(System.in);
+        Boolean exit = false;
```
yyixue: maybe this boolean variables can be named to sound more like a boolean?

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
+        } else {
+            printlnWithIndent("You must specify a event time after /at");
+        }
+    }
+
+    // Create a todo task
+    protected static void todo(Vector<Task> tasks, String[] arguments) {
+        String description = String.join(" ", Arrays.copyOfRange(arguments, 1, arguments.length));
+        tasks.add(new ToDo(description));
+        printNewTask(tasks);
```
gcksean: Can consider renaming these methods as the naming does not really signify what the method does

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
+            case "event":
+                event(tasks, arguments);
+                break;
+            case "todo":
+                todo(tasks, arguments);
+                break;
+            default:
+                save(tasks, line);
+            }
+
```
yyixue: is this blank line here needed?

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
+                Task task = tasks.get(index - 1);
+                task.markAsDone();
+                tasks.set(index - 1, task);
+
+                printlnWithIndent("Nice! I've marked this task as done:");
+                printlnWithIndent("\t" + task);
+            } catch (NumberFormatException e) {
+                printlnWithIndent("Index provided is not a proper number.");
+            } catch (IllegalArgumentException e) {
+                printlnWithIndent("Task with this index is not found in our database.");
```
yyixue: is this an if-else-catch block? not sure if there is supposed to be a catch at the end of the if-else block

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
+        for (int i = 0; i < tasks.size(); i += 1) {
+            printlnWithIndent(String.format("%d.\t%s", i + 1, tasks.get(i)));
+        }
+    }
+
+    // Mark a task to be done with index specified in arguments[1]
+    protected static void done(Vector<Task> tasks, String[] arguments) {
+        if (arguments.length < 2) {
+            // An index must be provided for the task to be marked "done"
+            printlnWithIndent("You will need to give me an index, like this: `done 2`.");
```
yyixue: i like how you added error handling for input of the index

src/main/java/Duke.java
```diff
@@ -1,10 +1,166 @@
...
+            }
+        }
+    }
+
+    // Create a deadline task
+    protected static void deadline(Vector<Task> tasks, String[] arguments) {
+        int i = findIndex(arguments, "/by");
+        if (i != -1) {
+            String description = String.join(" ", Arrays.copyOfRange(arguments, 1, i));
+            String by = String.join(" ", Arrays.copyOfRange(arguments, i + 1, arguments.length));
```
gcksean: Is the use of String.join() necessary here? Unless I'm missing something, it is possible to achieve the same result by just indexing the arguments array

src/main/java/Duke.java
fsgmhoward: Thanks for this advice. Will change in a new commit.

src/main/java/Duke.java
fsgmhoward: This two catches  should not have a relation with the if-else. Exception thrown in the try block (no matter whether they are in if block or not) will be caught by these two catches.

src/main/java/Duke.java
fsgmhoward: Thank you for the advice. I have renamed them as `*Handler` in the latest commit. Each of these are handler for a specific command so `*Handler` should be self-explanatory enough.

src/main/java/Duke.java
fsgmhoward: Thank you for the advice. I wanted to use `String.join` and `Arrays.copyOfRange` to split the tokenized arguments (and also remove the 1st element of the tokens as that's a command like `todo`, `deadline` etc.). May I know if there is a better solution for this?

You mentioned by just indexing the arguments array, do you mean something like this:
```
String by = "";
for (int j = i + 1; j < arguments.length; ++j) {
    by += arguments[j] + " ";
}
by = by.trim();
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,59 @@
...
+        ui.close();
+    }
+
+    public static void main(String[] args) {
+        String filepath = DEFAULT_SAVE_PATH;
+        if (args.length > 0) {
+            // If additional argument is provided, take 1st argument as the save filepath
+            filepath = args[0];
+        }
+        new Duke(filepath).run();
```
xhxh96, line 64: Good job in keeping your main method clean!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.lang.reflect.Constructor;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+import duke.command.Command;
+import duke.exception.InvalidInputException;
+import duke.exception.InvalidInputException.InputExceptionType;
+
+public class Parser {
```
xhxh96, line 15: You might want to include a Javadoc comment explaining what this class does

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,75 @@
...
+        } catch (ClassNotFoundException e) {
+            throw new SaveException(SaveExceptionType.INVALID_SAVE, e);
+        } finally {
+            // Close all open file and object handles
+            if (os != null) {
+                os.close();
+            }
+            if (fs != null) {
+                fs.close();
+            }
```
xhxh96: You can consider SLAP-ing this for better reusability since you're using it in the following `save` method as well

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,75 @@
...
+
+            // Nullify storage attribute before saving
+            tasks.setStorage(null);
+            os.writeObject(tasks);
+            tasks.setStorage(this);
+        } catch (FileNotFoundException e) {
+            // Thrown by `new FileOutputStream`, usually when `file` is a folder or cannot be created
+            throw new SaveException(SaveExceptionType.INVALID_PATH);
+        } finally {
+            // Close all open file and object handles
```
xhxh96: Good use of inline comments to describe the logic

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,40 @@
...
+        setStorage(storage);
+    }
+
+    public void setStorage(Storage storage) {
+        this.storage = storage;
+    }
+
+    /*
+     * Save file on all the modification actions
+     */
```
xhxh96: You might want to move this to a Javadoc comment for the class itself

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public void print(String text) {
+        String[] lines = text.split("\n");
+        for (String line : lines) {
+            System.out.println(indent + line);
+        }
+    }
+
+    // Print a message for a successful insertion of task
```
xhxh96: Use Javadoc comment to describe the purpose of a method

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public void printException(Exception e) {
+        print("Oops! " + e.getMessage());
+    }
+
+    public void printWelcome() {
+        print(
+                "Hello! I'm Duke.\n"
+                + "What can I do for you?"
```
xhxh96: Avoid magic strings -- consider declaring this as a constant

src/main/java/duke/command/ByeCommand.java
```diff
@@ -0,0 +1,19 @@
+package duke.command;
+
+import java.util.HashMap;
+
+import duke.TaskList;
+import duke.Ui;
+
+// Print out a bye message and do necessary clean up
```
xhxh96: Use Javadoc comment to describe the purpose of a class

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,41 @@
...
+        super(ui, tasks, arguments);
+    }
+
+    @Override
+    public void execute() throws InvalidInputException, IOException, SaveException {
+        String indexString = arguments.get("payload");
+        if (indexString == null) {
+            // An index must be provided for the task to be marked "done"
+            throw new InvalidInputException(InputExceptionType.EMPTY_INDEX);
+        } else {
```
xhxh96: You can consider removing this else clause so that there's lesser tendency for your code to be deeply nested

src/main/java/duke/exception/InvalidInputException.java
```diff
@@ -0,0 +1,43 @@
...
+public class InvalidInputException extends Exception {
+    public static enum InputExceptionType {
+        UNKNOWN_COMMAND,
+        EMPTY_DESCRIPTION,
+        EMPTY_INDEX,
+        NO_BY_DATE,
+        NO_AT_DATE,
+        NOT_INTEGER,
+        INDEX_OUT_OF_BOUND
+    }
```
xhxh96: Good use of enumerations!

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,21 @@
+package duke.task;
+
+import duke.exception.InvalidInputException;
+
+public class Deadline extends Task {
+    // This value indicates the version of this class when doing serialization
+    public static final long serialVersionUID = 01L;
+
+    protected String by;
```
xhxh96: You might want to replace this with a more descriptive name since `by` can be understood as either the person who created the task or the deadline of the task

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,47 @@
...
+        if (description == null || description.isEmpty()) {
+            throw new InvalidInputException(InputExceptionType.EMPTY_DESCRIPTION);
+        }
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        // Return tick when done, or 'X' symbol when it is not done
+        return (isDone ? "\u2713" : "\u2718");
```
xhxh96: You'd probably want to declare these unicode characters as a constant with descriptive names

xhxh96 (commented):
Overall good attempt at applying OOP. Pay attention to the use of magic string and use of appropriate comments. Otherwise code is clean and readable. Keep up the good work!

src/main/java/duke/command/DeleteCommand.java
fsgmhoward: Thanks for the advice. This has been fixed in the latest commit.

src/main/java/duke/Storage.java
fsgmhoward: They are actually different as this is the input stream and the `save` ones are output stream. Probably because of the inappropriate naming. It has been renamed more clearly in the latest commit.

src/main/java/duke/Ui.java
fsgmhoward: This has been moved to a separate class and declared as constant in latest codes. Thank you for the advice.

src/main/java/duke/task/Deadline.java
fsgmhoward: Thanks for the advice. This has been fixed in the latest commit.

src/main/java/duke/command/ByeCommand.java
fsgmhoward: Thanks for the advice. This has been fixed in the latest commit.

src/main/java/duke/Ui.java
fsgmhoward: Thanks for the advice. This has been fixed in the latest commit.

src/main/java/duke/TaskList.java
fsgmhoward: Thanks for the advice. This has been fixed in the latest commit.

src/main/java/duke/Parser.java
fsgmhoward: Thanks for the advice. This has been fixed in the latest commit.

PR #22, opened 2021-01-29
[Zufiqqar] iP

Zufiqqar:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
```
kwokyto: I like how clean this class is, there are no redundant functions here. Also, no coding violations here. Well done! (:

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
+        Task[] userTasks = new Task[MAX_NO_OF_TASKS];
+        int taskCounter = 0;
+        boolean exitLoopStatus = false;
+        Scanner inputCommand = new Scanner(System.in);
+        String userInput;
+
+        while(true){
+            userInput = inputCommand.nextLine();
+            String[] individualWords = userInput.split(" ", 2);
+            switch(individualWords[0].toLowerCase()){
```
kwokyto: This is a correct formatting for switch, good job for keeping that in mind!

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
+    private static boolean terminateProgram() {
+        boolean exitLoopStatus;
+        System.out.println("Bye. Hope to see you again soon!");
+        exitLoopStatus = true;
+        return exitLoopStatus;
+    }
+
+    private static void markActivityAsDone(Task[] userTasks, String[] individualWords) {
+        int activityNumber;
+        try {
```
kwokyto: This is the correct try-catch layout, good job!

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
+            case "done":
+                markActivityAsDone(userTasks, individualWords);
+                break;
+            case "bye":
+                exitLoopStatus = terminateProgram();
+                break;
+            default:
+                System.out.println("Oops. Unknown command. Try again");
+            }
+            if (checkLoopStatus(exitLoopStatus)) {
```
kwokyto: This is the correct if-conditional layout with the 'break' in a separate line, well done!

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
+            userInput = inputCommand.nextLine();
+            String[] individualWords = userInput.split(" ", 2);
+            switch(individualWords[0].toLowerCase()){
+            case "list":
+                displayListOfActivities(userTasks, taskCounter);
+                break;
+            case "todo":
+                taskCounter = createTodoTask(userTasks, taskCounter, individualWords);
+                break;
+            case "deadline":
```
kangxinwang: Good job in making code readable!

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
+import  java.util.Scanner;
+
 public class Duke {
+
+    static final int MAX_NO_OF_TASKS = 100;
```
kangxinwang: Good job in using a named constant!

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
     public static void main(String[] args) {
+        showWelcomeMessage();
+
+        Task[] userTasks = new Task[MAX_NO_OF_TASKS];
+        int taskCounter = 0;
+        boolean exitLoopStatus = false;
+        Scanner inputCommand = new Scanner(System.in);
+        String userInput;
+
+        while(true){
```
kwokyto: Should this while loop be extracted out? I think it might be cleaner if abstracted

src/main/java/Duke.java
```diff
@@ -1,10 +1,125 @@
...
+            default:
+                System.out.println("Oops. Unknown command. Try again");
+            }
+            if (checkLoopStatus(exitLoopStatus)) {
+                break;
+            }
+        }
+    }
+
+    private static boolean checkLoopStatus(boolean exitLoopStatus) {
```
kangxinwang: Maybe can rename the method to sound more like a verb?

src/main/java/Duke.java
kwokyto: Good job with the internal abstraction, very neat!

src/main/java/Duke.java
kwokyto: I am not sure if you should do the try{function} catch error instead of having the try catch in the function itself. perhaps it would be better to try the function instead

src/main/java/Task.java
```diff
@@ -0,0 +1,44 @@
...
+        this.description = description;
+        this.isDone = false;
+        this.taskType = taskType;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "\u2713" : "\u2718"); //return tick or X symbols
+    }
+
+    public String toString() {
```
kwokyto: I think you could add in "@Override" above this?

src/main/java/duke/app/Duke.java
```diff
@@ -0,0 +1,165 @@
...
+        Scanner inputCommand = new Scanner(System.in);
+        String userInput;
+        FileOperation.readFile(userTasks);
+
+        while(true){
+            userInput = inputCommand.nextLine();
+            String[] individualWords = userInput.split(" ", 2);
+            switch(individualWords[0].toLowerCase()){
+            case "list":
+                displayListOfActivities(userTasks);
```
Mick609: You use two different words (activity and task) to describe the tasks in this program. Maybe it is better to be consistent with one.

src/main/java/duke/app/Duke.java
```diff
@@ -0,0 +1,165 @@
...
+            case "deadline":
+                createDeadlineTask(userTasks, individualWords);
+                FileOperation.writeFile(userTasks);
+                break;
+            case "event":
+                createEventTask(userTasks, individualWords);
+                FileOperation.writeFile(userTasks);
+                break;
+            case "done":
+                markActivityAsDone(userTasks, individualWords);
```
Mick609: Same issue: if you choose to use the word task, maybe this method should be called "markTaskAsDone".

src/main/java/duke/file/FileOperation.java
```diff
@@ -0,0 +1,120 @@
+package duke.file;
+
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.Todo;
+
+import java.io.*;
```
Mick609: You should probably avoid using wildcard import like "import java.io.*".

src/main/java/duke/file/FileOperation.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.StringTokenizer;
+
+public class FileOperation {
+    public static String relativeFileName = "duke.txt";
+    public static final String SEPARATOR = "|";
+
+    public static void readFile(ArrayList<Task> userTasks){
+        char taskSelection;
+        try {
+            File myObj = new File(relativeFileName);
```
Mick609: You should probably avoid naming the intermediate variable to something like "myObj". Try to give it a name that explains its usage.

src/main/java/duke/file/FileOperation.java
```diff
@@ -0,0 +1,120 @@
...
+                    break;
+                case 'd':
+                case 'D':
+                    newLineToBeInserted = convertDeadlineDetailsToString(userTask);
+                    break;
+                case 'e':
+                case 'E':
+                    newLineToBeInserted = convertEventDetailsToString(userTask);
+                    break;
+                }
```
Mick609: You should probably add a default case for the switch statement.

Mick609 (commented):
The script is well written though there are some minor issues with the naming of variables.

PR #25, opened 2021-01-30
[Stefanie Low] iP

s-t-e-f:


src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
+        String description = "";
+        if (userInput.substring(0,8).equalsIgnoreCase("deadline")){
+            String by = "";
+            userInput = userInput.substring(9).trim();
+            if (userInput.contains("/")){
+                int idx = userInput.indexOf('/');
+                description =userInput.substring(0, idx);
+                by = userInput.substring(idx+3).trim();
+            }
+            else {
```
hiongkaihan: I think that for if else statement, we are supposed to flush the else after the if statement closing bracket }. For example,

if (condition) {
    statements;
} else {
    statements;
}

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
+import java.util.ArrayList;
+
 public class Duke {
+
+    static ArrayList<Task> tasks = new ArrayList<> ();
+    static int tasksCount = 0 ;
+
+    public static void printTaskList(ArrayList<Task> tasks) {
+        System.out.println("Here are the tasks in your list: ");
+        for (int i=0; i<tasks.size(); i++){
```
hiongkaihan: Good to standardize having a space between ) and {

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
     public static void main(String[] args) {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        System.out.println("------------------------------------------");
```
8kdesign: Perhaps you can combine the strings so that you only need to use one println

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
+    static int tasksCount = 0 ;
+
+    public static void printTaskList(ArrayList<Task> tasks) {
+        System.out.println("Here are the tasks in your list: ");
+        for (int i=0; i<tasks.size(); i++){
+            System.out.println("    " + (i+1) + ". " + tasks.get(i));
+        }
+    }
+
+    public static void addTask(String userInput) {
```
8kdesign: Perhaps you can refactor this method based on the different types of tasks to make it shorter

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
+        String description = "";
+        if (userInput.substring(0,8).equalsIgnoreCase("deadline")){
```
8kdesign: Perhaps you can consider using ".startsWith" so that you won't have magic numbers
Same for the other if statements below

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
 public class Duke {
+
+    static ArrayList<Task> tasks = new ArrayList<> ();
+    static int tasksCount = 0 ;
+
+    public static void printTaskList(ArrayList<Task> tasks) {
```
8kdesign: Perhaps you can remove tasks as a parameter
The method should be able to access the arraylist even without having it as a parameter

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
+            if (userInput.contains("/")){
+                int idx = userInput.indexOf('/');
+                description =userInput.substring(0, idx);
+                by = userInput.substring(idx+3).trim();
+            }
+            else {
+                description =userInput;
+            }
+
+            Deadline d = new Deadline(description, by);
```
8kdesign: Perhaps you can give it a name like "deadline" instead of just 'd'

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
+            if (task.equalsIgnoreCase("bye")) {
+                toContinueAddingTask = false;
+                break;
+            }
+            else if (task.equalsIgnoreCase("list")) {
+                printTaskList(tasks);
+                System.out.println("------------------------------------------");
+            }
+            else if (task.contains("done")){
+                int idx = Integer.parseInt(String.valueOf(task.charAt(5)));
```
8kdesign: Might want to declare a constant for the value 5 with a name that matches its purpose

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+
+    String by = "";
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    public String toString() {
+        return String.format("[D][ ] " + description + " (by: " + by + ")");
```
8kdesign: Can I check if isDone status is implemented for printing of deadline and events?
It seems like the "getStatusIcon" method is only used in Todo

8kdesign (commented):
Overall, I found the code readable, though there are some Java coding standard violations that you might want to address.

src/main/java/Duke.java
```diff
@@ -1,10 +1,105 @@
...
+            else {
+                description =userInput;
+            }
+
+            Deadline d = new Deadline(description, by);
+            tasks.add(d);
+        }
+        else if (userInput.substring(0,5).equalsIgnoreCase("event")){
+            String at = "";
+            userInput = userInput.substring(6).trim();
```
hiongkaihan: Maybe can use constants instead of magic numbers

hiongkaihan (commented):
Good job with the project! Overall code was well written with minor inconsistencies. I think it would help if you added some comments for some of the more complicated functions!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,37 @@
...
+    public void run() {
+        ui.showWelcome();
+
+        boolean isExit = false;
+        while (!isExit) {
+            String fullCommand = ui.readCommand();
+            if (fullCommand.equalsIgnoreCase("bye")) {
+                isExit = true;
+                Ui.printExitMessage();
+                Storage.saveFile();
```
wamikamalik: You can consider implementing SLAP here.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,105 @@
+package duke;
+
+
+import exceptions.DukeException;
+import tasks.*;
```
wamikamalik, line 4: you can replace the wildcard import with list of class you are using here. According to our coding standards, imported classes should always be listed explicitly. See [Coding Standards](https://se-education.org/guides/conventions/java/basic.html#statements)

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,105 @@
...
+            Ui.printTaskList();
+        }
+        else if (task.contains("done")) {
+            TaskList.markTaskDone(task);
+        }
+        else if (task.contains("delete")) {
+            TaskList.deleteTask(task);
+        }
+        else if (task.contains("find")) {
+            Ui.printMatchingTasks(task);
```
wamikamalik: You can consider declaring the magic strings (done, delete, find, list) with constants. Similar changes can be made in other places as well.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,105 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class Parser {
+    static final int TODOLENGTH = 4;
+    static final int EVENTLENGTH = 5;
+    static final int DEADLINELENGTH = 8;
+    static final int DELETELENGTH = 6;
```
wamikamalik: Consider adding an underscore to differentiate between one word and the next. For example, 
`private static final int TODO_LENGTH = 4`

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public static Event processAddEvent(String userInput) throws DukeException {
+        if (userInput.length() == EVENTLENGTH) {
+            throw new DukeException(TaskType.EVENT);
+        }
+        String description, at;
+        userInput = userInput.substring(EVENTLENGTH+1).trim();
+        if (userInput.contains("/")){
+            int idx = userInput.indexOf('/');
+            description =userInput.substring(0, idx);
```
wamikamalik: Can consider spacing out the operators consistently.
For example, `description = userInput.substring(0, idx);`
Similar spacing issues in other places as well.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+        String description = userInput.substring(TODOLENGTH+1);
+        return new Todo(description);
+    }
+
+    public static LocalDate processDatesTimes(String dateString) {
+        LocalDate date = null;
+        Pattern patt = Pattern.compile("2[0-9][0-9][0-9]-[0-1][0-9]-[0-3][0-9]");
+        Matcher matcher = patt.matcher(dateString);
+        boolean matchFound = matcher.find();
```
wamikamalik: You could consider renaming this variable so that it sounds more like a boolean variable.

PR #45, opened 2021-01-31
[Emily] iP

EmilyTJX:


src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
...
+    protected String deadline;
+
+    public Deadline(String description, String deadline) {
+        super(description);
+        this.deadline = deadline;
+    }
+
+    @Override
+    public String toString(){
+        return "[D]" + super.toString() + "(by: " + deadline + ")";
```
huachen24: Should [D] be a constant in the class so that it is clearer what it means? Like each subclass of Task can have a taskType?

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
+
+
 public class Duke {
+    private static final String divider = "_______________________________\n";
```
PingruiLi: A constant name should be in capital letter

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+        tasks.add(currentTask);
+        System.out.println(divider + "Got it. I've added this task:\n" +
+                currentTask.toString() + "\n");
+        System.out.println("Now you have "+tasks.size()+" tasks in your list.\n" + divider);
+    }
+
+    public static void addDeadline(String input){
+        int deadlineIndex = input.indexOf("/");
+        String description = input.substring(9, deadlineIndex);
+        String deadline = input.substring(deadlineIndex+4);
```
PingruiLi: the naming here is a little confusing, it might be better if you use something like "doBy" or "finishBy".

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+            }else if(input.contains("deadline")){
+                addDeadline(input);
+            }else if(input.contains("event")){
+                addEvent(input);
+            }else if(input.equals("bye")){
+                exitDuke();
+                break;
+            }else{
+                addTask(input);
+            }
```
PingruiLi, line 54: well done! the branch is in correct format and  correctly follows SLAP

src/main/java/Task.java
```diff
@@ -0,0 +1,20 @@
...
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "\u2718" : " "); //if done return X
```
PingruiLi: it might be better if declaring "\u2718" as a string constant at the beginning. although you add comment here, it looks like a magical literal.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task{
+    protected String deadline;
```
PingruiLi: the naming is a little confusing as the member variable has the same name as the class

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+
+    public static void addEvent(String input){
+        int timeIndex = input.indexOf("/");
+        String description = input.substring(6, timeIndex);
+        String timeDetails = input.substring(timeIndex + 4);
+        Task currentTask = new Event(description, timeDetails);
+        tasks.add(currentTask);
+        System.out.println(divider + "Got it. I've added this task:\n" +
+                currentTask.toString() + "\n");
+        System.out.println("Now you have " + tasks.size() + " tasks in your list.\n" + divider);
```
PingruiLi: the print message part is repeated in addTodo, addDeadline and addEvent, maybe you could consider making it a separate method.

PingruiLi (commented):
Your code is well done but there are minor things that could still be improved.

src/main/java/Duke.java
huachen24: I think that divider can be named in all caps, since it is a constant.

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+    private static void markTaskDone(int index){
+        Task currentTask = tasks.get(index-1);
+        currentTask.markAsDone();
+        System.out.println(divider + "Nice! I've marked this task as done:"+
+                "\n"+ currentTask.toString() +
+                "\n" + divider);
+    }
+
+    public static void addToDo(String input){
+        String taskName = input.substring(5);
```
huachen24: 5 seems like a magical number here. Would it be better if you change it to get the index of the start of the task name?

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+        Task currentTask = new ToDo(taskName);
+        tasks.add(currentTask);
+        System.out.println(divider + "Got it. I've added this task:\n" +
+                currentTask.toString() + "\n");
+        System.out.println("Now you have "+tasks.size()+" tasks in your list.\n" + divider);
+    }
+
+    public static void addDeadline(String input){
+        int deadlineIndex = input.indexOf("/");
+        String description = input.substring(9, deadlineIndex);
```
huachen24: Same as addToDo, not sure how 9 is derived without knowing the command syntax

src/main/java/Duke.java
huachen24: Could the 4 possibly change if the user enters a typo?

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+        Task currentTask = new Deadline(description, deadline);
+        tasks.add(currentTask);
+        System.out.println(divider + "Got it. I've added this task:\n" +
+                currentTask.toString() + "\n");
+        System.out.println("Now you have "+tasks.size()+" tasks in your list.\n");
+    }
+
+    public static void addEvent(String input){
+        int timeIndex = input.indexOf("/");
+        String description = input.substring(6, timeIndex);
```
huachen24: Same as addToDo

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+
+    }
+
+    private static void exitDuke(){
+        System.out.println("Bye. Hope to see you again soon!");
+    }
+
+
     public static void main(String[] args) {
         String logo = " ____        _        \n"
```
huachen24: I think that this can be all caps as well since it is a constant

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+
+        System.out.println("Hello! I'm Duke\nWhat can I do for you?\n");
+        Scanner in = new Scanner(System.in);
+        String input;
+
+        while (in.hasNext()) {
+            input = in.nextLine();
+            if (input.equals("list")) {
+                listTask();
+            }else if(input.contains("done")) {
```
huachen24: When using contains(), would there be a case where `todo get homework done` triggers this and thus create an exception where 'get homework done' cannot be converted to an int?

src/main/java/Event.java
```diff
@@ -0,0 +1,13 @@
...
+    protected String eventTime;
+
+    public Event(String description, String eventTime) {
+        super(description);
+        this.eventTime = eventTime;
+    }
+
+    @Override
+    public String toString(){
+        return "[E]" + super.toString() + "(at: " + eventTime + ")";
```
huachen24: Same as Deadline class

src/main/java/Task.java
huachen24: If you're just returning an X, would it be better to just use 'X' so that the confusing unicode can be avoided?

src/main/java/Task.java
```diff
@@ -0,0 +1,20 @@
...
+    }
+    public void markAsDone(){
+        this.isDone = true;
+    }
+    public String toString() {
+        return "[" + this.getStatusIcon() + "] " + this.description;
+    }
+}
```
huachen24: I think that it would be good to standardize having a newline at the end of file for all files

src/main/java/ToDo.java
```diff
@@ -0,0 +1,11 @@
+public class ToDo extends Task{
+
+    public ToDo(String description) {
+        super(description);
+    }
+
+    @Override
```
huachen24: I like that you added the Override tag so that it's immediately clear that this method is being overridden

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
+
+
```
huachen24, line 5: I think that 2 blank lines is enough between import statement and the rest of the code

huachen24 (commented):
I think overall the code conforms to the coding standards. Good job! :)

src/main/java/DukeCommandHandler.java
```diff
@@ -0,0 +1,184 @@
...
+        for (int i = 0; i < tasks.size(); ++i) {
+            System.out.println("\t" + (i + 1) + "." +
+                    tasks.get(i).toString());
+        }
+        System.out.print(DIVIDER);
+    }
+
+    protected static void markTaskDone(String input) {
+        //task number can be found on 5th index of input
+        int taskNumber = Integer.parseInt(input.substring(5));
```
sevenseasofbri: Perhaps avoid the usage of a magic number here? Consider declaring a constant to store this value. (Referring the the number '5')

src/main/java/DukeCommandHandler.java
```diff
@@ -0,0 +1,184 @@
...
+        Task currentTask = new Event(taskDescription, timeDetails);
+        tasks.add(currentTask);
+        System.out.println(DIVIDER + "\tGot it. I've added this task:\n\t\t" +
+                currentTask.toString());
+        System.out.print("\tNow you have " + tasks.size() + checkSingular() + "in your list.\n" + DIVIDER);
+
+    }
+
+    protected static void deleteTask(String input) {
+        //to add exception
```
sevenseasofbri: Perhaps avoid comments that are not directly addressed to the reader/ do not explain 'what' or 'why' of the code?

https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#write-to-the-reader

src/main/java/DukeCommandHandler.java
```diff
@@ -0,0 +1,184 @@
...
+            //if input after "to-Do" command is only whitespace
+            DukeException.toDoInvalidDescription();
+            return;
+        }
+
+        Task currentTask = new ToDo(taskInfo);
+        tasks.add(currentTask);
+        System.out.println(DIVIDER + "\tGot it. I've added this task:\n" +
+                "\t\t" + currentTask.toString());
+        System.out.print("\tNow you have " + tasks.size() + checkSingular() + "in your list." + "\n" + DIVIDER);
```
sevenseasofbri: Perhaps you can practice SLAP here by extracting this out into a method? Similarly, this can be done for other statements to be printed and create `printXYZ()` methods.

src/main/java/DukeCommandHandler.java
```diff
@@ -0,0 +1,184 @@
...
+        Task currentTask = tasks.get(taskNumber - 1);
+        tasks.remove(taskNumber-1);
+        System.out.print(DIVIDER + "\tNoted. I've removed this task:" +
+                "\n\t" + currentTask.toString() + "\n"
+                + DIVIDER);
+        System.out.print("\tNow you have " + (tasks.size()) + checkSingular() + "in your list.\n" + DIVIDER);
+
+    }
+
+    protected static void exitDuke() {
```
sevenseasofbri: The naming of this function seems to be a little off as it doesn't describe what the function really does. Consider re-naming it to `printExitMessage()` or something along those lines?

src/main/java/Duke.java
```diff
@@ -1,10 +1,52 @@
...
+    public Duke() {
+        // set up the things needed to start Duke
+        FileManager.createFolder();
+        storage = new FileManager();
+        if (storage.retrieveTextFile()) {
+            storage.loadData();
+        }
+    }
+
+    public void run() {
```
sevenseasofbri, line 40: Good use of SLAP in this function! I like that you extracted the handlers into methods. :)

src/main/java/DukeException.java
```diff
@@ -0,0 +1,40 @@
+public class DukeException {
```
sevenseasofbri: Since DukeException is an exception class, shouldn't it extend the Exception class? This might change a bit of the error handling part of the code since you haven't thrown DukeException, but called the functions from this class for all the error cases... Consider handling the exceptions with try, catch, throw instead of using the class to call the functions and then using `return;` ?
https://nus-cs2113-ay2021s2.github.io/website/schedule/week5/topics.html#c-to-java-exceptions-how-to-use-exceptions

sevenseasofbri (commented):
Overall, good use of SLAP and following coding standards!

One thing though, there are a lot of comments in the code which seem a little redundant because the code is sufficient to describe that is happening. Consider removing some of them.

PR #46, opened 2021-01-31
[Vikas Harlani] iP

vvvvh123:


src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+    private static final int EVENT_LENGTH = 6;
+    private static final int BY_AT_LENGTH = 3;
+    private static Task[] tasksList = new Task[100];
+    private static int taskCounter = 0;
+
     public static void main(String[] args) {
+        showWelcomeMessage();
+        inputLoop();
+    }
+
```
hazelhedmine: I like how you separated the sections in your code with a new line as the clear structure increases code readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    private static final int TODO_DONE_LENGTH = 5;
+    private static final int DEADLINE_LENGTH = 9;
+    private static final int EVENT_LENGTH = 6;
+    private static final int BY_AT_LENGTH = 3;
```
hazelhedmine: I like how you stored the length of the words as constants as it makes your main code easier to read.

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+    private static Task[] tasksList = new Task[100];
```
hazelhedmine: Should this be "tasks" instead of "tasksList" as it is representing a collection of objects?

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    private static final int TODO_DONE_LENGTH = 5;
```
hazelhedmine: Perhaps you could consider separating "TODO_DONE_LENGTH" to two separate constants of "TODO_LENGTH" and "DONE_LENGTH"? This may increase code readability. I noticed the similar naming style in line 8 of your code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+        System.out.println("\t  " + tasksList[taskCounter-1]);
+        if (taskCounter == 1) {
+            System.out.println("\tNow you have " + taskCounter + " task in the list.");
+        } else {
+            System.out.println("\tNow you have " + taskCounter + " tasks in the list.");
+        }
+        printLine();
+    }
+
+    private static void printLine() {
```
hazelhedmine: Perhaps you could name this as printHorizontalLine instead? This may increase your code readability since printLine may be misleading as in line 19 you named your user command as "line".

hazelhedmine (commented):
Overall, I felt like your code was well organized and readable! I've suggested minor changes to improve the readability of your code. The naming of your variables were clear but I noticed one possible coding violation which I've commented on. I also appreciated how you used functions to organize your code!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,18 @@
...
+    protected String by;
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String getStatusIcon() {
+        return (isDone ? "\u2713" : " ");
```
huachen24: Would it be a good idea to add a comment to explain the unicode so that people know what you are returning?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by + ")";
```
huachen24: To me it seems a little unintuitive as to what D might mean if it's my first time seeing the code, perhaps you could make it an attribute of the Deadline class?

src/main/java/Duke.java
huachen24: I agree with Hazel on this, the same could be done for BY_AT_LENGTH as well.

src/main/java/Duke.java
huachen24: To add on, List is an interface in Java so perhaps avoiding list would be a better idea since you are using an array in this case

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+                showExitMessage();
+                break;
+            } else if (line.equals("list")) {
+                printTasksList();
+                continue;
+            } else if (line.startsWith("done")) {
+                int index = Integer.parseInt(line.substring(TODO_DONE_LENGTH)) - 1;
+                tasksList[index].markAsDone();
+                printDoneTask(index);
+                continue;
```
huachen24: continue appears here but not in subsequent else if statements, perhaps you could standardize across the whole while loop?

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+            }
+            taskCounter++;
+            printNewTask();
+        }
+    }
+
+    private static void printDoneTask(int index) {
+        printLine();
+        System.out.println("\tNice! I've marked this task as done:");
+        System.out.print("\t" + "  ");
```
huachen24: I don't think there is a need for concatenation here, as "\t  " should work just fine

src/main/java/Duke.java
```diff
@@ -1,10 +1,107 @@
...
+            System.out.println(tasksList[i-1]);
+        }
+        printLine();
+    }
+
+    private static void printNewTask() {
+        printLine();
+        System.out.println("\tGot it. I've added this task:");
+        System.out.println("\t  " + tasksList[taskCounter-1]);
+        if (taskCounter == 1) {
```
huachen24: I like how u split the singular and plural cases

src/main/java/Duke.java
huachen24: I like that you have a separate method to printing the divider line as it is frequently used

src/main/java/Event.java
```diff
@@ -0,0 +1,18 @@
...
+    protected String at;
+
+    public Event(String description, String at) {
+        super(description);
+        this.at = at;
+    }
+
+    @Override
+    public String getStatusIcon() {
+        return (isDone ? "\u2713" : " ");
```
huachen24: Same as in the Deadline class and the rest of the classes

src/main/java/ToDo.java
```diff
@@ -0,0 +1,11 @@
+public class ToDo extends Task {
```
huachen24: I notice that ToDo doesn't override the getStatusIcon method from Task, whereas Deadline and Event does. Is this intended? And also since Task will never be called in your program, would it be better to just delete the overriding methods and just make the change in the method in the Task class?

huachen24 (commented):
Great job on this! I really like how you broke out certain methods and constants like printLine and the expected length of the commands. 
One small thing is that perhaps you want to add the .class files to your gitignore as they do not need to be pushed to the repo.

src/main/java/Duke.java
vvvvh123: thank you

src/main/java/Duke.java
vvvvh123: thank you

src/main/java/Duke.java
vvvvh123: thank you both, will update my code

src/main/java/Duke.java
vvvvh123: thank you both, will update my code

src/main/java/Duke.java
vvvvh123: thank you, will update my code

src/main/java/Deadline.java
vvvvh123: yes, definitely, will be adding comments, thank you!

src/main/java/Duke.java
vvvvh123: thank you, yes i will try to standardize it

src/main/java/ToDo.java
vvvvh123: Oh yes this was intended, because I felt like for deadlines and events, they shouldn't print a cross when not completed, only a tick when completed (because there is a time aspect for those).

src/main/java/Duke.java
vvvvh123: Thank you

src/main/java/Duke.java
vvvvh123: thank you

src/main/java/Duke.java
vvvvh123: oh true thank you

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,290 @@
...
+
+public class Duke {
+
+    private static final int TODO_LENGTH = 5;
+    private static final int DONE_LENGTH = 5;
+    private static final int DEADLINE_LENGTH = 9;
+    private static final int EVENT_LENGTH = 6;
+    private static final int BY_LENGTH = 3;
+    private static final int AT_LENGTH = 3;
+    private static final int DELETE_LENGTH = 7;
```
rafaelperes: great use of constants.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,290 @@
...
+
+    private static final String ROOT_PATH = System.getProperty("user.dir");
+    private static final Path FOLDER_PATH = Paths.get(ROOT_PATH, "data");
+    private static final Path FILE_PATH = Paths.get(ROOT_PATH, "data", "duke.txt");
+
+    public static void main(String[] args) {
+        createFile();
+        loadFile();
+        showWelcomeMessage();
+        inputLoop();
```
rafaelperes: good use of abstractions ;)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,290 @@
...
+        try {
+            List<String> lines = Files.readAllLines(FILE_PATH);
+            for (String line : lines) {
+                String[] components = line.split("/");
+                String taskType = components[0];
+                String done = components[1];
+                String description = components[2];
+                Task newTask = new Task(description);
+                switch (taskType) {
+                case "T":
```
rafaelperes: maybe use constants here (enum?) to reduce the need of magic strings and improve your code readability?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,290 @@
...
+                String description = t.getDescription();
+                String done = (t.isDone()) ? "1" : "0";
+                if (t instanceof ToDo) {
+                    fw.write("T" + "/" + done + "/" + description + System.lineSeparator());
+                } else if (t instanceof Deadline) {
+                    String by = ((Deadline) t).getBy();
+                    fw.write("D" + "/" + done + "/" + description + "/" + by + System.lineSeparator());
+                } else if (t instanceof Event) {
+                    String at = ((Event) t).getAt();
+                    fw.write("E" + "/" + done + "/" + description + "/" + at + System.lineSeparator());
```
rafaelperes: fw.write three times, maybe we need an abstraction for this?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,290 @@
...
+            System.out.println("\t" + exceptionGreeting + "I've encountered an error saving your data :-(");
+        }
+    }
+
+    private static void inputLoop() {
+        Scanner in = new Scanner(System.in);
+        String line;
+        while (true) {
+            line = in.nextLine();
+            if (line.equals("bye")) {
```
rafaelperes: refer to the magic string comments

src/main/java/duke/task/ToDo.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class ToDo extends Task {
+
+    public ToDo(String description) {
+        super(description);
+    }
+
+    @Override
+    public String toString() {
+        return "[T]" + super.toString();
```
rafaelperes, line 24: refer to the magic string comments

rafaelperes (commented):
Great job Vikas!

PR #48, opened 2021-01-31
[Alicia Tay] iP

aliciatay-zls:


src/main/java/Duke.java
```diff
@@ -1,10 +1,200 @@
...
+        System.out.println("\t- displays all tasks in the list" + "\n");
+        System.out.println("\ttodo     | <task>");
+        System.out.println("\t- adds specified task to the list" + "\n");
+        System.out.println("\tdeadline | <task>  | /by | <deadline>");
+        System.out.println("\t- adds specified task and deadline and to the list" + "\n");
+        System.out.println("\tevent    | <task>  | /at | <timing>");
+        System.out.println("\t- adds specified task and timing to the list" + "\n");
+        System.out.println("\tdone     | <index>");
+        System.out.println("\t- marks existing task matching the specified index as completed in the list" + "\n");
+        System.out.println("<> indicates an input field and | is a field separator");
```
fsgmhoward: I feel this can use `+` to connect strings to avoid multiple calls to System.out.println(), like this:
```
String helpString = "HELP PAGE" + NEWLINE
        + "All valid commands:" + NEWLINE + NEWLINE
        + ....
```
And you declared `NEWLINE` but you sometimes didn't use it (e.g. here)

src/main/java/Duke.java
```diff
@@ -1,10 +1,200 @@
...
+        try {
+            Integer.parseInt(tokens[1]);
+        } catch (NumberFormatException e) {
+            return false;
+        }
+        int index = Integer.parseInt(tokens[1]);
+        return index > 0 && index <= tasksCount;
+    }
+
+    public static boolean hasDeadlineOrEvent(String keyword, String[] tokens) {
```
fsgmhoward: Naming: `isDeadlineOrEvent` should be more appropriate as you are checking whether tokens match the pattern of deadline or event. Or alternatively, naming it as `hasKeyword` is also fine in my opinion.

src/main/java/Duke.java
```diff
@@ -1,10 +1,200 @@
...
+        String userInput = in.nextLine();
+        while (!userInput.equals("bye")) {
+            if (userInput.equals("list")) {
+                printList();
+            } else if (userInput.equals("help")) {
+                help();
+            } else {
+                processInput(userInput);
+            }
+            userInput = in.nextLine();
```
fsgmhoward: a single line user input is processed in different functions (different levels of abstraction) [REF](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#slap-hard)

src/main/java/Duke.java
```diff
@@ -1,10 +1,200 @@
...
+        echo();
+    }
+
+    public static void markInList(int id) {
+        tasks[id-1].setStatus(DONE_STATUS);
+        System.out.println(BORDER);
+        System.out.println("Nice! This task is now done:");
+        System.out.print("\t");
+        tasks[id-1].printTask();
+        System.out.print("\n");
```
fsgmhoward: Declared `NEWLINE` but not used here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,200 @@
...
+    private static final String BORDER = "____________________________________________________________";
+    private static final String NEWLINE = System.lineSeparator();
+    private static final String DONE_STATUS = "X";
+
+    public static void greet() {
+        System.out.println(BORDER);
+        help();
+        System.out.println("Hello, I'm Panda!");
+        System.out.println("What would you like to do today?");
+        System.out.println(BORDER + NEWLINE);
```
fsgmhoward: For each function there is repeat `System.out.println(BORDER);`
Maybe this can be moved to before Duke.c#L189 and after Duke.c#L196, since anyway these two will be printed each time an input is received?

fsgmhoward (commented):
Looks good! Just some suggestions given on code quality as commented below.

src/main/java/Duke.java
```diff
@@ -1,10 +1,239 @@
...
+        echo();
+    }
+
+    public static void markInList(int id) {
+        tasks[id-1].setStatus(DONE_STATUS);
+        System.out.println(BORDER);
+        System.out.println("Nice! This task is now done:");
+        System.out.print("\t");
+        tasks[id-1].printTask();
+        System.out.print("\n");
```
oscarlai1998: You have already declared Newline so you can use it here

oscarlai1998 (commented):
Hello, your code is extremely neat and easy to follow. You have consistently followed the coding standard which makes it hard for me to even spot a mistake. I love how you have customised your code and used a text art!

src/main/java/Duke.java
```diff
@@ -1,10 +1,239 @@
...
+        String userInput = in.nextLine();
+        while (!userInput.equals("bye")) {
+            if (userInput.equals("list")) {
+                printList();
+            } else if (userInput.equals("help")) {
+                help();
+            } else {
+                processInput(userInput);
+            }
+            userInput = in.nextLine();
```
FarmZH98: I like how you made the options very clear and straightforward.

src/main/java/Duke.java
```diff
@@ -1,10 +1,239 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    /** Constants used for displaying messages */
+    private static final String BORDER = "____________________________________________________________";
+    private static final String NEWLINE = System.lineSeparator();
+    public static final String LOGO =
```
FarmZH98: Well done in using all uppercase and underscore to separate words in constant names

src/main/java/Task.java
```diff
@@ -0,0 +1,41 @@
+public class Task {
+    private String item;
+    private int index;
+    private String status;
+    private String type;
```
FarmZH98: I think the namings are not entirely clear, for example can the "type" be named as "classType" instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,239 @@
...
+        default:
+            printInvalidInputMessage();
+            break;
+        }
+        if (!isValid) {
+            printInvalidInputMessage();
+        }
+    }
+
+    public static boolean hasIndex(String[] tokens) {
```
FarmZH98: I like how the function is named to sound like booleans.

FarmZH98 (commented):
Good job overall in complying with the coding standards I can think of. I had a tough time as well spotting violations.

src/main/java/Duke.java
aliciatay-zls: Thank you, I will. Thank you for the detailed suggestions as well!

src/main/java/Duke.java
aliciatay-zls: That's true, especially since my return statement uses isSomething instead of hasSomething type booleans. I'll change it to isDeadlineOrEvent. Thank you!

src/main/java/Duke.java
aliciatay-zls: Thank you for pointing this out, I will correct it in my next increment.

src/main/java/Duke.java
aliciatay-zls: You're right, thank you! Will do.

src/main/java/Task.java
aliciatay-zls: That makes more sense, I will edit this. Appreciate it!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,392 @@
...
+
+
+    /** Methods that display messages */
+    public static void greet() {
+        String greetString = BORDER + NEWLINE
+                + LOGO + NEWLINE
+                + "Hello, I'm Panda!" + NEWLINE
+                + "What would you like to do today?" + NEWLINE
+                + "Tip: use \"help\" to view all valid commands" + NEWLINE
+                + BORDER + NEWLINE + NEWLINE;
```
xhxh96: Consider declaring constants as uppercase SNAKE_CASE as per coding standard.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    public static void printInvalidInputMessage(String message) {
+        String openString = BORDER + NEWLINE;
+        String closeString = NEWLINE + BORDER + NEWLINE + NEWLINE;
+        if (message != null) {
+            System.out.print(openString + message + closeString);
+            return;
+        }
+        System.out.print(openString + DEFAULT_INVALID_MESSAGE + closeString);
```
xhxh96: Good use of guard clause! Although in this case guarding `message == null` since printing the message can be thought of as the happy path in this case.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,392 @@
...
+                + "as completed in the list" + NEWLINE  + NEWLINE
+                + "<> indicates an input field and | is a field separator." + NEWLINE
+                + BORDER + NEWLINE + NEWLINE;
+        System.out.print(helpString);
+    }
+
+
+    /** Methods that print part of or full list */
+    public static void echo() {
+        System.out.print(BORDER + NEWLINE);
```
xhxh96: Consider SLAP-ing this if you're reusing the same print line a couple of times.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,392 @@
...
+        System.out.print(tasksCount > 1 ? " tasks" : " task");
+        System.out.print(" in your list." + NEWLINE);
+        System.out.print(BORDER + NEWLINE + NEWLINE);
+    }
+
+    public static void printList() {
+        System.out.print(BORDER + NEWLINE);
+        System.out.print("Here are the tasks in your list:" + NEWLINE);
+        for (int i=0; i<tasksCount; i++) {
+            System.out.print("\t" + (i+1) + ". ");
```
xhxh96: Consider adding a space between operators for better readability

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,392 @@
...
+        printNumberOfTasks();
+    }
+
+    
+    /** Methods that check if user inputs are valid commands */
+    public static void processInput(String userInput) {
+        if (userInput.equals("list")) {
+            printList();
+            return;
+        } else if (userInput.equals("help")) {
```
xhxh96: You do not actually need an else in this case if you're using guard clause.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,392 @@
...
+                printInvalidInputMessage(INVALID_INDEX_MESSAGE);
+            } catch (DukeException e) {
+                printInvalidInputMessage(OUTSIDE_RANGE_INDEX_MESSAGE + tokens[1]);
+            }
+            break;
+        default:
+            printInvalidInputMessage(null);
+            break;
+        }
+    }
```
xhxh96: Consider SLAP-ing the individual case branches for better readability.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,392 @@
...
+        loadFromDisk();
+        Scanner in = new Scanner(System.in);
+        String userInput = in.nextLine();
+        while (!userInput.equals("bye")) {
+            processInput(userInput);
+            userInput = in.nextLine();
+        }
+        String textToOverwrite = convertListToText();
+        saveToDisk(textToOverwrite);
+        goodbye();
```
xhxh96: Good job in keeping the main method clean!

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public void printTask() {
+        System.out.print("[" + this.getType() + "] "
+                + "[" + this.getStatus() + "] "
+                + this.getItem());
+    }
+
+    @Override
+    public String toString() {
+        int statusAsInt = this.getStatus() == "X" ? 1 : 0;
```
xhxh96: Avoid magic numbers. In this case you'd probably want to declare what 1 and 0 mean.

xhxh96 (commented):
Overall good naming of variables and use of guard clauses. Code is generally clean. Well done!

PR #54, opened 2021-01-31
[Hazel Hedmine Tan] iP

hazelhedmine:


src/main/java/Duke.java
```diff
@@ -31,49 +32,49 @@ public static void main(String[] args) {
...
+                System.out.println(HORIZONTAL_LINE + "\n Nice! I've marked this task as done:");
+                System.out.println("   " + task.toString()  + "\n" + HORIZONTAL_LINE);
+                break;
+            case "deadline":
+                splitCommand = command.split(" ", 2);
+                date = splitCommand[1].split("/by", 2);
+                Deadline deadline = new Deadline(date[0].trim(), date[1].trim());
+                tasks[numberOfTasks] = deadline;
+                numberOfTasks++;
+                System.out.println(" Alright, I've added this task:\n   " + deadline.toString() + "\n"
```
nivikcivik: Hey since you repeat the "Alright, I've added this task..." statement for all 3 types of tasks, maybe you want to create a separate method like "printAddedTask" and print the statements from there? you can also update the value of numberOfTasks within the method as well since it is common for all 3 types of tasks:)

src/main/java/Duke.java
```diff
@@ -31,49 +32,49 @@ public static void main(String[] args) {
...
+            case "event":
+                splitCommand = command.split(" ", 2);
+                date = splitCommand[1].split("/at", 2);
+                Event event = new Event(date[0].trim(), date[1].trim());
+                tasks[numberOfTasks] = event;
+                numberOfTasks++;
+                System.out.println(" Alright, I've added this task:\n   " + event.toString() + "\n"
+                        + " Now you have " + numberOfTasks + " tasks in your list.\n" + HORIZONTAL_LINE);
+                break;
+            case "todo":
```
nivikcivik: since the "case "todo": " and "default" have the same contents, maybe you can keep just one of them? this would make it shorter and more readable in my opinion:) but if you have different meanings for the two of them that's fine, but generally I cannot find any code violations:))

src/main/java/Duke.java
```diff
@@ -4,18 +4,19 @@
 
     public static void main(String[] args) {
         Scanner userInput = new Scanner(System.in);
-        String command;
         String HORIZONTAL_LINE = "____________________________________________________________";
```
nivikcivik: perhaps this could be in small caps? you could use camelCase convention:)

src/main/java/Duke.java
```diff
@@ -4,18 +4,19 @@
...
         String LOGO = " ____        _        \n"
```
nivikcivik: I think "LOGO" can also be in small caps:)

src/main/java/Duke.java
```diff
@@ -31,49 +32,49 @@ public static void main(String[] args) {
...
-                    toDo = new ToDo(command);
-                    tasks[numberOfTasks] = toDo;
-                    numberOfTasks++;
-                    System.out.println(" Alright, I've added this task:\n   " + toDo.toString() + "\n"
-                            + " Now you have " + numberOfTasks + " tasks in your list.\n" + HORIZONTAL_LINE);
-                    break;
-                }
+                continue;
+            }
+            String[] tokens = command.split(" ");
```
nivikcivik: maybe instead of tokens you can use something like "partsOfCommand"? I think it can be more descriptive, but otherwise I see no other issues:) your code is so neat and well-written:)

nivikcivik (commented):
Overall your code is very well written! It is very neat and readable. There are very few changes required. Otherwise your coding style is very consistent:)

src/main/java/Duke.java
hazelhedmine: Thank you for the suggestion! I think it will improve my code readability hence will likely incorporate that in soon :)

src/main/java/Duke.java
hazelhedmine: Thank you for noting this, I will make sure to rectify that soon!

src/main/java/Duke.java
hazelhedmine: I was unsure if this should be in all capital or camelCase convention, thank you for pointing this out to me! I will change it in my next commit :)

src/main/java/Duke.java
hazelhedmine: Noted, I will incorporate that in! Thank you for your suggestions :)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+import java.util.ArrayList;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
+public class Duke {
+
+    public static void main(String[] args) {
```
sevenseasofbri: Consider shortening this method, it it currently spanning lines 21 to 130 (109 lines!). Perhaps one way of doing this is to extract related sections of code into methods and use SLAP. 
https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+
+        Scanner userInput = new Scanner(System.in);
+
+        printStartingMessage();
+
+        ArrayList<Task> tasks = new ArrayList<>();
+        Task task;
+        String[] description;
+        String inputLine;
+        int COMMAND_TASK_SEPARATOR = 2;
```
sevenseasofbri: Since `int COMMAND_TASK_SEPARATOR` is a constant, consider declaring it as `static` and `final`?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+        String filePath = new File("").getAbsolutePath();
+
+        try {
+            tasks = loadfromFile(filePath + "/duke.txt");
+        } catch (FileNotFoundException e) {
+            System.out.println(" I've created a text file at " + filePath + " to save your tasks!");
+            printHorizontalLine();
+        }
+
+        while (!(inputLine = userInput.nextLine().trim()).equals("bye")) {
```
sevenseasofbri: In order to increase readability, perhaps avoid assigning the value for `inputLine` inside the `while()` conditional.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+                try {
+                    int doneTaskNumber = Integer.parseInt(command[1]);
+                    task = tasks.get(doneTaskNumber - 1);
+                    task.markAsDone();
+                    printDoneMessage(task);
+                } catch (IndexOutOfBoundsException e) {
+                    printNonExistentTaskMessage();
+                } catch (NumberFormatException e) {
+                    printNotANumberMessage();
+                }
```
sevenseasofbri: The two cases have varying levels of abstraction, consider standardizing the level to SLAP. For example, the contents of the `done` case can be extracted to a separate method to mark the task as done. Perhaps the same could be done for the other case statements as well? This could reduce the size of the `main` method considerably.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+                    ToDo toDo = new ToDo(command[1]);
+                    tasks.add(toDo);
+                    printAddedMessage(tasks, toDo);
+                } catch (ArrayIndexOutOfBoundsException e) {
+                    printMissingFieldsMessage();
+                }
+                break;
+            case "delete":
+                try {
+                    int deleteTaskNumber = Integer.parseInt(command[1]);
```
sevenseasofbri: Consider renaming `deleteTaskNumber` to a noun (like `taskNumberToBeDeleted`) rather than an action word.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+                fw.write(line + "\n");
+            }
+            fw.close();
+        } catch (IOException e) {
+                System.out.println("ERROR: something went wrong! :(");
+        }
+    }
+
+    public static ArrayList<Task> loadfromFile(String filePath) throws FileNotFoundException {
+        //creates a file for filepath
```
sevenseasofbri: Consider removing this comment? It seems a bit redundant.

sevenseasofbri (commented):
Overall, your code follows a good amount of coding standards and conventions. The naming is well done! I like that you separated the printing operations into multiple functions. I have provided my comments in the review. Good job!

src/main/java/duke/Duke.java
hazelhedmine: Thank you for letting me know, I will implement this soon!

src/main/java/duke/Duke.java
hazelhedmine: I was unsure about this part, thank you for clarifying and I will make sure to change it!

PR #55, opened 2021-01-31
[Mark Low Seng Keong] iP

marklowsk:


src/main/java/Duke.java
```diff
@@ -1,10 +1,313 @@
...
+    }
+
+    /**
+     * Prints a statement with a tab and spacing character.
+     * Can print multiple lines if the statement string has line
+     * breaks in it.
+     */
+    private static void printStatement(String statement) {
+        String[] fragments = statement.split("\\R");
+        for (String fragment : fragments) {
```
xseh: Iterator variables can be called i, j, k etc. In this case, use f instead of fragment.

src/main/java/Duke.java
```diff
@@ -1,10 +1,313 @@
...
+            return;
+        }
+        recordTask(new Event(taskDescription, eventAt));
+    }
+
+    /**
+     * Records the given task into the global task array.
+     * Increments the global task count.
+     *
+     * @param task task object to be recorded
```
xseh: The comments should end with a full stop.

src/main/java/Duke.java
```diff
@@ -1,10 +1,313 @@
...
+        case COMMAND_DEADLINE_WORD:
+            recordDeadline(commandArgs);
+            break;
+        case COMMAND_EVENT_WORD:
+            recordEvent(commandArgs);
+            break;
+        case COMMAND_DONE_WORD:
+            markTaskDone(commandArgs);
+            break;
+        case COMMAND_BYE_WORD:
```
xseh: Add // Fallthrough comment for cases that does not have break.

src/main/java/Duke.java
```diff
@@ -1,10 +1,313 @@
...
+    private static final String ERROR_INVALID_EVENT_MESSAGE = "Invalid event.";
+    private static final String ERROR_PREFIX_MESSAGE = "ERROR: ";
+    private static final String ERROR_INVALID_TASK_NAME_MESSAGE = "Invalid task name.";
+    private static final String ERROR_INVALID_TASK_NUMBER_MESSAGE = "Invalid task number.";
+
+    private static final int MAX_NUMBER_OF_TASKS = 100;
+    private static Task[] tasks = new Task[MAX_NUMBER_OF_TASKS];
+    private static int numberOfTasks = 0;
+
+    private static final Scanner SCANNER = new Scanner(System.in);
```
kewenlok: Scope of a variable should be minimised. Since scanner is only used in one part of the code, it is best to make it a local variable rather than a global one.

src/main/java/Duke.java
```diff
@@ -1,10 +1,313 @@
...
+    /**
+     * Parses an argument value after a given token.
+     * Reads until a next token or the end of commandArgs string.
+     * If token is null, read from the start of the string until
+     * a next token or the end of commandArgs string.
+     *
+     * @param commandArgs a full string of command arguments
+     * @param token a string representing an option portion of an argument e.g. "/by"
+     * @return the argument value after the given token
+     */
```
kewenlok: I like the fact that you have gone ahead of the class and created javadoc for most of the methods, it lets me follow through the code easily.

src/main/java/Duke.java
```diff
@@ -1,10 +1,313 @@
...
+    private static void markTaskDone(String commandArgs) {
+        try {
+            String argValue = parseArgument(commandArgs, null);
+            if (argValue == null) {
+                throw new Exception(ERROR_INVALID_TASK_NUMBER_MESSAGE);
+            }
+            int taskNumber = Integer.parseInt(argValue);
+            if (taskNumber > 0 && taskNumber <= numberOfTasks) {
+                int taskIndex = taskNumber - 1;
+                tasks[taskIndex].isDone = true;
```
kewenlok: You might want to implement a setDone setter method in the task class to change the isDone variable. The class level variable should be accessed and modified by setter and getter method within the class.

src/main/java/Duke.java
```diff
@@ -1,10 +1,313 @@
...
+    private static String parseArgument(String commandArgs, String token) {
+        int readFromIndex = 0;
+        int readUntilIndex = commandArgs.length();
+        if (token != null) {
+            if (commandArgs.contains(token)) {
+                readFromIndex = commandArgs.indexOf(token) + token.length();
+            } else {
+                return null;
+            }
+        }
```
kewenlok: I think it is better to separate the conditional statements into multiple individual ones and make the main code of assigning the readFromIndex value more prominent for easier management and readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,313 @@
...
+        printStatements(EXIT_MESSAGE);
+        System.exit(0);
+    }
+
+    /**
+     * Parses the user input and attempts to execute the command
+     * with the arguments.
+     */
+    private static void executeCommand(String userInput) {
+        String[] parsedCommand = parseCommand(userInput);
```
xseh: Plural form should be used on names representing a collection of objects.

xseh (commented):
Some minor coding standard violation but overall very clear and clean code. Good job!

src/main/java/Duke.java
marklowsk: Thanks for pointing out, totally forgot to implement the setter.

src/main/java/Duke.java
marklowsk: thanks!

src/main/java/Duke.java
marklowsk: Thanks for the advice!

src/main/java/Duke.java
marklowsk: I will modify the variable name, thanks!

src/main/java/Duke.java
marklowsk: Thanks for pointing that out!

src/main/java/Duke.java
marklowsk: Thanks for the info, will place it within scope.

src/main/java/duke/commands/CommandParser.java
```diff
@@ -0,0 +1,59 @@
+package duke.commands;
+
+import duke.data.exceptions.InvalidCommandException;
+
+public class CommandParser {
+    /**
+     * Parses the user input into its command word and argument (if exists),
```
okkhoy, line 11: Good use of javadoc comment here. 👍

src/main/java/duke/commands/CommandParser.java
```diff
@@ -0,0 +1,59 @@
...
+     * If the command word is invalid, throw InvalidCommandException.
+     *
+     * @param userInput input from the user.
+     * @return a Command object to be executed by caller.
+     * @throws InvalidCommandException If command word does not exist.
+     * @see #parseCommand(String)
+     */
+    public static Command parse(String userInput) throws InvalidCommandException {
+        String[] commandAndArgs = parseCommand(userInput);
+        String commandName = commandAndArgs[0];
```
okkhoy: I do see some magic numbers here!

src/main/java/duke/commands/CommandParser.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Splits the user input into a command word and command arguments string.
+     * Returns an array containing the command word and arguments string.
+     *
+     * @return an array of size 2; first element is the command word
+     *         and second element is the arguments string.
+     */
+    private static String[] parseCommand(String userInput) {
+        final String[] commandAndArgs = userInput.trim().split("\\s+", 2);
```
okkhoy: Just saying (i.e., in this case may not be much damage): this line is particularly a low-level operation as compared to the rest of this method. So SLAP is weak

src/main/java/duke/commands/DoneCommand.java
```diff
@@ -0,0 +1,43 @@
...
+            int taskNumber = parseNumberFromArgument(commandArgs);
+            tasks.markTaskDone(taskNumber, ui, storage);
+        } catch (DukeException e) {
+            ui.printError(e.getMessage());
+        } catch (NumberFormatException e) {
+            ui.printError(ERROR_NOT_A_TASK_NUMBER_MESSAGE);
+        }
+    }
+
+    public void execute(TaskList tasks, TextUI ui, Storage storage) {
```
okkhoy, line 43: should these be marked as `@override` for better understandability?

src/main/java/duke/common/Utils.java
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @param dateInput a string containing the date supplied from the user.
+     * @return a LocalDateTime object or null if failed to parse dateInput.
+     */
+    public static LocalDateTime parseDate(String dateInput) {
+        for (DateTimeFormatter dtf : POSSIBLE_DATE_FORMATS) {
+            try {
+                return LocalDate.parse(dateInput, dtf).atStartOfDay();
+            } catch (DateTimeParseException e) {
+                // Used to catch errors from parsing date.
```
okkhoy: it would be better to throw an exception rather than have an empty catch block

src/main/java/duke/data/exceptions/InvalidCommandException.java
```diff
@@ -0,0 +1,11 @@
+package duke.data.exceptions;
+
```
okkhoy, line 2: These group of exception classes could use some Javadoc comments to inform the other developers what they are meant to do.

src/main/java/duke/data/task/TaskList.java
```diff
@@ -0,0 +1,258 @@
...
+     *
+     * @param task task object to be recorded
+     * @param ui the TextUI object that that engages user input and program output.
+     * @param storage the Storage object that writes/retrieves to/from file.
+     */
+    public void recordTask(Task task, TextUI ui, Storage storage) {
+        tasks.add(task);
+        ui.printStatements(TASK_ADDED_MESSAGE,
+                String.format(DOUBLE_SPACE_PREFIX_STRING_FORMAT, task),
+                String.format(TASK_TOTAL_TASKS_STRING_FORMAT, tasks.size()));
```
okkhoy: Do you think the constants could be named better?

okkhoy (commented):
In general, very neat code. Good job; please keep it up!

src/main/java/duke/commands/DoneCommand.java
marklowsk: Thanks prof for pointing that out!

src/main/java/duke/data/exceptions/InvalidCommandException.java
marklowsk: Good suggestion, thanks!

src/main/java/duke/commands/CommandParser.java
marklowsk: 👍 thanks!

src/main/java/duke/common/Utils.java
marklowsk: Thanks prof for pointing that out!

src/main/java/duke/commands/CommandParser.java
marklowsk: Thanks prof for pointing that out!

src/main/java/duke/commands/CommandParser.java
marklowsk: Ok good point, I combined this method back.

src/main/java/duke/data/task/TaskList.java
marklowsk: Yes just realized it is unpleasant to read hahaha

PR #58, opened 2021-02-01
[Gerard Tan Wei Kiat] IP

gerardtwk:


src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
+import java.util.Scanner;
+
 public class Duke {
+    private static final String DIVIDER = "____________________________________________________________";
```
blasco-git: I like that you have used a constant to store the Divider. You can try implementing a method to print the Divider. In that way, you will only need to call the method instead of typing System.out.println(Divider)

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
                 + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        System.out.println(DIVIDER);
+        System.out.println(logo);
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        System.out.println(DIVIDER);
+    }
+
+    public static void handleCommand(String command) {
```
blasco-git: I like that you only used abstractions for this method. It makes it easy to read and understand.

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
+import java.util.Scanner;
+
 public class Duke {
+    private static final String DIVIDER = "____________________________________________________________";
+    private static final Task[] STORED_TASKS = new Task[100];
```
blasco-git: Good job! You did not violate any coding standard.

blasco-git: I think you should consider using a constant to declare the array. It will improve your code quality.

blasco-git (commented):
Good job overall! Your code is neat, readable and understandable. Keep up the good work!

src/main/java/Duke.java
```diff
@@ -1,10 +1,112 @@
...
+        do {
+            command = scanner.nextLine();
+            handleCommand(command);
+        } while(!command.equals("bye"));
+
+        scanner.close();
+    }
+
+    public static void welcomeMsg() {
         String logo = " ____        _        \n"
```
Emkay16: Since the Duke logo is unlikely to change while in use, would it perhaps be better to declare this as a constant?

Emkay16 (commented):
Code is very clear and readable, could not find any major coding convention violations 👍 
I like how everything is so neat, keep up the great work!

src/main/java/Duke.java
```diff
@@ -1,10 +1,278 @@
...
+                    }
+                } else {
+                    currentTask = new Event(taskInfo[2], taskInfo[3]);
+                    if (taskInfo[1].equals("1")) {
+                        currentTask.markAsDone();
+                    }
+                }
+                STORED_TASKS.add(currentTask);
+            }
+            saveFileScanner.close();
```
okkhoy: Sort of an indicator that SLAP may be violated here. You can try to extract out the stuff inside the if...else ladder to another method.
Also how `taskInfo[1]` is treated could be confusing to the one who embarks on maintaining your code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,278 @@
...
+        } catch (Exception e) {
+            e.printStackTrace();
```
okkhoy: try not to do this for exception handling. the user wouldn't know what to make out of a stack trace

src/main/java/Duke.java
```diff
@@ -1,10 +1,278 @@
...
+            System.out.println(DIVIDER);
+            System.out.println("oops, something went wrong...");
+            System.out.println(DIVIDER);
+        }
+    }
+
+    public static void saveDataToFile() {
```
okkhoy: try to add header comments for non-trivial public methods

src/main/java/Duke.java
```diff
@@ -1,10 +1,278 @@
...
+    }
+
+    public static void deleteTask(String command) throws DukeException {
+        if (command.equals("delete")) {
+            throw new DukeException("☹ OOPS!!! The task number of a delete command cannot be empty.");
+        }
+
+        if (!command.startsWith("delete ")) {
+            throw new DukeException("☹ OOPS!!! The correct syntax for delete command is: 'delete task_number'");
+        }
```
okkhoy: Can these checks be done elsewhere?
*you may want to create more classes/methods to enable checking elsewhere.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,5 @@
+public class DukeException extends Exception {
```
okkhoy, line 1: You could describe the exception and its purpose using header comments.

okkhoy (commented):
Please pay attention to SLAP and try to improve SLAP where possible.
Next, please add some header comments to your code.

PR #70, opened 2021-02-01
 iP

violinyap:


src/main/java/Duke.java
```diff
@@ -1,10 +1,90 @@
...
+        System.out.println("____________________________________________________________");
+
+        Scanner scanner = new Scanner(System.in);
+        String text = "";
+        Boolean isContinue = true;
+
+        Todo[] todos = new Todo[100];
+        Integer index = 0;
+
+        while (isContinue) {
```
Mick609: How does the while loop end and where do you use the variable "isContinue"?

src/main/java/Duke.java
```diff
@@ -1,10 +1,90 @@
...
+                    System.out.println("Sorry, I do not know what you mean.");
+                }
+            }
+
+        }
+
+        scanner.close();
+
     }
 }
```
Mick609, line 52: The main method contains almost all the operations of the program. Please consider using methods for SLAP.

src/main/java/Duke.java
```diff
@@ -1,10 +1,90 @@
...
+                }
+                System.out.println("____________________________________________________________");
+
+            } else {
+                String[] arr = text.split(" ", 2);
+                String action = arr[0];
+                String input = arr[1];
+
+                if (action.equals("done")){
+                    int setAsDone =Integer.parseInt(input) ;
```
Mick609: I think the int variable here is for the index of the task that is about to be set as done. Using "setAsDone" as the name of the int variable is a bit misleading.

src/main/java/Duke.java
```diff
@@ -1,10 +1,90 @@
...
+        System.out.println("____________________________________________________________");
+        System.out.println("Hello! I'm \n" + logo );
+        System.out.println("What can I do for you?");
+        System.out.println("____________________________________________________________");
+
+        Scanner scanner = new Scanner(System.in);
+        String text = "";
+        Boolean isContinue = true;
+
+        Todo[] todos = new Todo[100];
```
Mick609: In this program, you use Todo as the parent class for Deadline and Event. I found this a bit misleading as the todo, event, and deadline should be three tasks of the same level (all children of the parent class Task). This is just a suggestion.

src/main/java/Duke.java
```diff
@@ -1,10 +1,90 @@
...
+                    todos[index++] = newTodo;
+
+                    System.out.println("____________________________________________________________");
+                    System.out.println("Got it. I've added this task: ");
+                    System.out.println(todos[index - 1]);
+                    System.out.println("Now you have " + index + " tasks in the list.");
+                    System.out.println("____________________________________________________________");
+
+                } else if (action.equals("deadline")) {
+                    String[] inputs = input.split(" /by ", 2);
```
Mick609: You should probably consider using a different name for the product of the split method. Maybe "splittedInputs"?

Mick609 (commented):
I have noticed the project here did not complete several increments (i.e., level 4 and level 5) required. There are some naming issues for variables and the code does not have comments. Meanwhile, most operations are done within the main method. Please consider using methods for SLAP.

PR #72, opened 2021-02-02
[Soe Paing Zaw] iP

soepaingzaw:


src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
+import java.util.Scanner;
```
ongweisheng, line 5: Firstly, i think you can have more commits to better show your updates and improve on your commit name/message

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
     public static void main(String[] args) {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
+
         System.out.println("Hello from\n" + logo);
+
+        for (int i = 0; i < LINES; i++) {
```
ongweisheng: Perharps you can just create the string for the horizontal line using a method instead of a for loop to print our _ 25 times

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+        int num;
+        int temp;
+
+        while(true) {
+
+            tasks[index] = scan.nextLine();
+            checks[index] = "[ ]";
+            parts = tasks[index].split(" ");//splits string into separate parts
+
+            if(parts[0].equals("done")) {
```
ongweisheng: Maybe use switch case to split up your task and use methods to handle the task required to keep the code clean

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
...
+                            type[num]+ checks[num] + " " + tasks[num]
+                    +deadline[num]);
+                    count--;
+                    num++;
+                }
+            }
+
+            //Below code increments index as tasks are being added
+            else {
+
```
ongweisheng, line 108: Same comment as above so as to avoid confusing nested loops and improve code clarity

src/main/java/Duke.java
```diff
@@ -1,10 +1,135 @@
+import java.util.Scanner;
+
```
ongweisheng, line 7: Secondly, not using classes as recommended for the iP for Task, Todo, Deadline, Event etc.

ongweisheng (commented):
Code quality guidelines review

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
 public class Duke {
+
+    static final int ITEMS = 25;
+    static final int LINES = 85;
+
     public static void main(String[] args) {
         String logo = " ____        _        \n"
```
andyrobert3: if logo does not change, can declar it as a constant

src/main/java/Duke.java
andyrobert3: agree, name it `printIntroText()`

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
+        System.out.println("What can I  do for you?");
+
+        for (int i = 0; i < LINES; i++) {
+            System.out.print("_");
+        }
+
+        System.out.print("\n");
+        Scanner scan = new Scanner(System.in);
+
+        //new String[numRowsofData]
```
andyrobert3: delete

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
+        String[] tasks = new String[ITEMS];
+        String[] checks = new String[ITEMS];
+        String[] type = new String[ITEMS];
+        String[] deadline = new String[ITEMS];
+        String[] parts;
+
+        int index = 0;
+        int count;
+        int num;
+        int temp;
```
andyrobert3: unclear variable namings

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
+                count = index;
+                num = 0;
+
+                while (count != 0) {
+
+                    System.out.println(num + 1 + ". " +
+                            type[num] + checks[num] + " " + tasks[num]
+                            + deadline[num]);
+                    count--;
+                    num++;
```
andyrobert3: Maybe can make this cleearer

src/main/java/Duke.java
```diff
@@ -1,10 +1,144 @@
...
+                System.out.println("Got it. I've added this task:\n  "
+                        + type[index] + checks[index] + " "
+                        + tasks[index] + " " + deadline[index++] + "\n"
+                        + "Now you have " + index + " task(s) in the list");
+
+                for (int i = 0; i < LINES; i++) {
+                    System.out.print("_");
+                }
+
+                System.out.print("\n");
```
andyrobert3: maybe a helper print function would help here

PR #75, opened 2021-02-02
[Wonjae Choi] iP

wjchoi0712:


src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
+        String userInput = in.nextLine();
+
+        while (!userInput.equals("bye")) {
+            if (userInput.startsWith("todo") || userInput.startsWith("deadline") || userInput.startsWith("event")) {
+                addTask(tasks, userInput);
+                taskCounter++;
+                printResponse(tasks, MESSAGE_ADDTASK, taskCounter);
+            } else if (userInput.equals("list")) {
+                printResponse(tasks, MESSAGE_LIST);
+            } else if (userInput.startsWith("done")) {
```
rageqqq: Using boolean variables instead of .startsWith and .equals can make things easier to understand.

src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
+    }
+
+    /**
+     * Prints out the full list of tasks with line boundaries surrounding it
+     *
+     * @param tasks list of tasks to be printed out
+     * @param response message that is shown to the user
+     *
+     */
+    public static void printResponse(ArrayList<Task> tasks, String response) {
```
rageqqq: Same method name as line 58, maybe use printList for this method

src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
+     */
+    private static void manageTasks() {
+        ArrayList<Task> tasks = new ArrayList<Task>();
+        int taskCounter = 0;
+
+        Scanner in = new Scanner(System.in);
+        String userInput = in.nextLine();
+
+        while (!userInput.equals("bye")) {
+            if (userInput.startsWith("todo") || userInput.startsWith("deadline") || userInput.startsWith("event")) {
```
rageqqq: Line can seem overly complicated, maybe use boolean variables to make it easier to digest

src/main/java/Duke.java
rageqqq: Storing "bye", "todo" etc. as strings can help when you need to add more code later on

rageqqq (commented):
Overall pretty easy to understand the code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
+    }
+
+    /**
+     * Separates the userInput to description and time string
+     *
+     * @param description userInput to be separated
+     * @return inputs description and time is stored in the string array
+     *
+     */
+    public static String[] separateDescriptionAndTime(String description) {
```
Vinci-Hu: It's quite good to introduce a method for the separation operation so that dealing with input is divided into small tasks.

src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
+                printResponse(tasks, MESSAGE_ADDTASK, taskCounter);
+            } else if (userInput.equals("list")) {
+                printResponse(tasks, MESSAGE_LIST);
+            } else if (userInput.startsWith("done")) {
+                int completedTaskNo = Integer.parseInt(userInput.replace("done ", ""));
+                tasks.get(completedTaskNo - 1).markAsDone();
+                printResponse(tasks, MESSAGE_DONETASK, completedTaskNo);
+            } else {
+                tasks.add(new Task(userInput));
+                taskCounter++;
```
Vinci-Hu: I have a doubt here. Should taskCounter be maintained here? What if the user did not type in a correct format like without a "/by " phrase, but it is still counted.

src/main/java/Duke.java
Vinci-Hu: I think the expressions here are pretty straightforward, only that they are not in the same format. What I mean is they don't look like parallel statements. These recognition for keywords part might be optimized later on.

src/main/java/ToDo.java
```diff
@@ -0,0 +1,9 @@
+public class ToDo extends Task {
+    public ToDo(String description) {
+        super(description);
+    }
+    @Override
+    public String toString() {
+        return "[T]" + super.toString();
+    }
```
Vinci-Hu: Very neat for calling super toString method.

src/main/java/Duke.java
```diff
@@ -1,10 +1,141 @@
...
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+                + "|____/ \\__,_|_|\\_\\___|\n\n";
+    private static final String MESSAGE_GREETING = " Hello! I'm Duke" + "\n What can I do for you?";
+    private static final String MESSAGE_ENDING = " Bye. Hope to see you again soon!";
+    private static final String MESSAGE_LIST = " Here are the tasks in your list:";
+    private static final String MESSAGE_ADDTASK = " Got it. I've added this task:";
+    private static final String MESSAGE_REMAININGTASK = " Now you have %d tasks in the list.";
+    private static final String MESSAGE_DONETASK = "Nice! I've marked this task as done:";
+
```
Vinci-Hu: The constant string above are very readable. This is good.

src/main/java/Duke.java
Vinci-Hu: It is function overloading, I think is quite straightforward. But can also separate into "printList" and "printResponse" as they are quite different as in the incode- comments.

Vinci-Hu (commented):
Code is neat and satisfies OO standards. Maybe the method names can be changed to make the meaning more straighforward and reader-friendly. The comment lines are helpful.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,245 @@
+package duke;
+
+import java.util.Scanner;
+import java.util.ArrayList;
+import duke.exception.*;
+import duke.task.*;
+import java.io.*;
```
stephlewyh: It would be good to explicitly state what classes would be used and maintained, like in the [import statement convention](https://se-education.org/guides/conventions/java/basic.html#package-and-import-statements).

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,245 @@
...
+        File f = new File(filePath);
+        Scanner s = new Scanner(f);
+        while (s.hasNext()) {
+            String data = s.nextLine();
+            char taskType = data.charAt(1);
+            String description = data.substring(7);
+            String detail;
+            String time;
+            switch (taskType) {
+            case 'T':
```
stephlewyh: Single characters may not be too useful in explaining what the task types actually are. See ['using name to explain'](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#use-name-to-explain) for suggestions.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,245 @@
...
+                tasks.add(new ToDo(description));
+                break;
+            case 'D':
+                detail = description.substring(0, description.indexOf("by:") - 2);
+                time = description.substring(description.indexOf("by:") + 4, description.length() - 1);
+                tasks.add(new Deadline(detail, time));
```
stephlewyh: There seems to be some [magic numbers](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#use-name-to-explain), yet not, as in line 229 to 230. Perhaps you would like to consider how the code could be kept [simple](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#practice-kissing), while giving meaning.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,17 @@
+package duke.task;
+
+import duke.task.Task;
+
+public class Event extends Task {
+    protected String at;
```
stephlewyh: This naming can suggest a location or time property. You could try another name that could increasing meaning here. See [https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#not-too-long-not-too-short](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#not-too-long-not-too-short).

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,245 @@
...
+                switch (command) {
+                case BYE:
+                    runDuke = false;
+                    break;
+                case LIST:
+                    printList(tasks);
+                    break;
+                case DONE:
+                    int completedTaskNo = Integer.parseInt(userInput.replace("done ", ""));
+                    tasks.get(completedTaskNo - 1).markAsDone();
```
stephlewyh: This line seems to be doing two things at once, which is [undesirable](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-complicated-expressions). You could consider adding another variable to keep count instead.

PR #77, opened 2021-02-02
[Jonathan Khoo] iP

JonathanKhooTY:


src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+            Task t = new Task(text);
+            if(text.equalsIgnoreCase("List")){
+                listArray(tasks);
+            }
+            else if(text.startsWith("Done") || text.startsWith("done")){
+                Integer taskIndex = Integer.parseInt(text.substring(5));
+                markAsDone(taskIndex);
+
+            }
+            else{
```
Zufiqqar: What you can do here is }else{ in 1 line if you really want to follow the coding standard

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
+import javax.sound.midi.SysexMessage;
+import java.util.Scanner;
+
 public class Duke {
+
+    static Task[] tasks = new Task[100];
+    static int taskPosition = 0;
+
+
+    public static void storeTask(Task t){
```
Zufiqqar: Easy to understand what this function does by the function name. Good use of the naming. I think to improve on the naming and readability further, the parameter t can be renamed to maybe newTask. This can be easily understood that the function storeTask is storing the newly created task into a certain position on the tasks array. Lastly, you will be using newTask.description instead of t.description.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,26 @@
...
+    public String getStatusIcon() {
+        return "[D]" + "[" + super.getStatusIcon() +"]";
+    }
+
+
+    @Override
+    public String getDescription() {
+        return super.getDescription() + " " + "(" + by + ")";
+    }
+
```
stephlewyh: These  lines do not serve a special purpose, and you could consider to remove them in favour of [code readability](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#delete-dead-code).

src/main/java/Deadline.java
```diff
@@ -0,0 +1,26 @@
+import javax.sound.midi.SysexMessage;
+
+public class Deadline extends Task {
+
+    protected String by;
+
+    public Deadline(String description, String by){
```
stephlewyh: Though it doesn't affect your runtime execution, it is preferred to maintain layout consistency in terms of keeping a space before the  `{` here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,139 @@
...
+            Scanner in = new Scanner(System.in);
+            text = in.nextLine();
+            //Task t = new Task(text);
+            String[] arr = text.split(" "); //split command input into words
+            String command = arr[0];
+
+            switch (command){
+            case "deadline":
+            case "Deadline":
+            case "DEADLINE":
```
stephlewyh: These seem to refer to the same case, except that the string inputs are different. You could consider a separate way of handling such string inputs before executing the deadline block.

src/main/java/Duke.java
```diff
@@ -1,10 +1,139 @@
...
+
+    public static void main(String[] args) throws InvalidCommandException, InvalidDateException{
+        System.out.println("Hello! I'm 343 Guilty Spark! You may call me Spark!");
+        System.out.println("What can I do for you?");
+        String text = "hi";
+
+        while(!text.equalsIgnoreCase("Bye")){
+            Scanner in = new Scanner(System.in);
+            text = in.nextLine();
+            //Task t = new Task(text);
```
stephlewyh: You could consider removing such lines as it is not a real comment. If this code could become useful again, you can refer to [this](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#delete-dead-code) to handle such code.

src/main/java/Event.java
```diff
@@ -0,0 +1,19 @@
+public class Event extends Task{
+
+    protected String by;
+    public Event(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+
```
stephlewyh: Again, spare lines like  these may not be necessary as you have a space between the methods already.

src/main/java/Task.java
```diff
@@ -0,0 +1,23 @@
+public class Task {
```
stephlewyh: Good job in keep this class simple and easy to read.

src/main/java/Duke.java
```diff
@@ -1,10 +1,260 @@
...
+
+    private static void restoreFileContents(String filePath) throws FileNotFoundException, BlankDescriptionException {
+        File f = new File(filePath);
+        Scanner s = new Scanner(f);
+        while (s.hasNext()) {
+            inputToTasks(s.nextLine());
+        }
+    }
+
+    private static void inputToTasks(String nextLine) throws BlankDescriptionException {
```
stephlewyh: This seems like a long method. Maybe you can try to abstract some of the details away, while maintaining [SLAP](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#slap-hard).

src/main/java/Duke.java
```diff
@@ -1,10 +1,260 @@
...
+        int taskStatusIndex = 1;
+        int taskDescriptionIndex = 7;
+
+        char taskType = nextLine.charAt(1);
+        String taskStatus = nextLine.substring(4,5);
+        String taskDescription = nextLine.substring(7);
+        try {
+            switch (taskType) {
+            case 'T':
```
stephlewyh: Naming a task type as a single character is [ambiguous](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#not-too-long-not-too-short).  How about trying other names that could directly describe the task types?

PR #85, opened 2021-02-02
[Chuah Jin Wei] iP

heyjinwei:


src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
...
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    public void markAsDone() {
```
zihan9485: Should setters for boolean variables be in the form setDone()?

src/main/java/Event.java
```diff
@@ -0,0 +1,14 @@
...
+
+    protected String at;
+
+    public Event(String description, String at) {
+        super(description);
+        this.at = at;
+        System.out.print("Got it. I've added this task:\n" + this + '\n');
+    }
+
+    public String toString() {
```
zihan9485: Perhaps consider including @Override when overriding the toString method. This could apply for the other classes as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
+import java.util.Scanner;
+import java.util.Random;
+
 public class Duke {
+
+    public static void printSeparator() {
```
zihan9485: I like the use of an extracted method to print the separator!

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
+        while (!line.equals("bye")){
+            lineParts = line.split(" ");
+            switch(lineParts[0]) {
+            case "list":
+                printTasks(tasks, taskCount);
+                break;
+            case "done":
+                int index = Integer.parseInt(lineParts[1]) - 1;
```
zihan9485: Perhaps a clearer name for index like taskIndex would be better?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+public class Deadline extends Task {
+
+    protected String by;
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+        System.out.print("Got it. I've added this task:\n" + this + '\n');
```
vaiish371: Any reason why you have not included ".by" with 'this' keyword? I think it would be better practice to specify which attribute the calling object is accessing and not to just directly access this object.

src/main/java/Duke.java
```diff
@@ -1,10 +1,84 @@
...
+        System.out.print('\n');
+    }
+
+    public static void exitMethod() {
+        System.out.print("My cover's blown!\n");
+        printSeparator();
+    }
+
+    public static void echoMode(String line) {
+        Random rd = new Random();
```
vaiish371: Using "rd" as the Random object name seems to be vague and might not go by the naming convention. Instead you can use a more specific word like randomNumber that provides better readability and explanation to what you are intending to do.

vaiish371 (commented):
I like how your code is simple and neat, and does not have any over-complicated components. It has a good flow from function to function, that makes it easier for readers to understand what the program does. Other than few naming conventions that you might've forgotten to pay heed to, it looks good to merge!

src/main/java/Task.java
heyjinwei: Good suggestion! The initial idea was to make it clear that the method only sets the task as done and not the other way. Would consider allowing the user to change it the other way as well!

src/main/java/Event.java
heyjinwei: Thank you! Would add it to my code.

src/main/java/Duke.java
heyjinwei: Thank you!

src/main/java/Duke.java
heyjinwei: Changed. Thank you!

src/main/java/Deadline.java
heyjinwei: I actually want it to print the .toString value every time. Will be adding .toString behind them.

src/main/java/Duke.java
heyjinwei: rd is not a random number, it can be used to generate random integers or floats too, depending on the method that is called. in this case, I called a .nextBoolean method. Not really sure what to name this random object

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,252 @@
...
+import java.util.Random;
+import java.io.FileWriter;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.io.IOException;
+import java.io.File;
+import duke.task.Task;
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Todo;
```
stephlewyh: Good job with these import statements that specify explicitly, the classes that are to be maintained here.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,252 @@
...
+
+public class Duke {
+
+    private static final ArrayList<Task> tasks = new ArrayList<>();
+    private static final String FILE_SEPARATOR = "CHOPCHOP";
+
+    private static void printSeparator() {
+        for(int i = 0; i < 60; i++) {
```
stephlewyh: 60 seems to be a [magic number](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-magic-numbers) that can be avoided.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,252 @@
...
+
+    /**
+     * Echoes the user input with random upper and lower case for mockery.
+     *
+     * @param line User input.
+     */
+    private static void mockEcho(String line) {
+        Random rd = new Random();
+        for(int i = 0; i < line.length(); i++) {
+            if(rd.nextBoolean()) {
```
stephlewyh: You could leave a space before `(` as in line 51 ,to maintain layout consistency.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,252 @@
...
+        System.out.print("I can remember your tasks for you!\n\n" +
+                "Available commands:\n" +
+                "\ttodo <description>\n" +
+                "\tdeadline <description> /by <time due>\n" +
+                "\tevent <description> /at <time occurring>\n" +
+                "\tlist\n" + "\tdone <task index>\n" +
+                "\tdelete <task index>\n");
+    }
+
+    private static void listMode() {
```
stephlewyh: This method is [long](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods), which could afford some revision. Perhaps abstract some details away?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,252 @@
...
+                            addTask(new Event(lineParts[2], lineParts[3], Boolean.parseBoolean(lineParts[1])));
+                        } catch (IndexOutOfBoundsException e) {
+                            System.out.print("Something went wrong. Please put the event time after /at.\n");
+                        } catch (DukeException e) {
+                            System.out.print("The description of a event cannot be empty.\n");
+                        }
+                        break;
+                    }
+                }
+            }
```
stephlewyh: There are a couple of indentations here, which could obstruct reading flow. See [https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent) for suggested improvements.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,252 @@
...
+            } else {
+                fw.write(t.getTaskType() + FILE_SEPARATOR + t.getDone() + FILE_SEPARATOR
+                        + t.getDescription() + FILE_SEPARATOR  + t.getDate() + "\n");
+            }
+        }
+        fw.close();
+    }
+
+    public static void main(String[] args) {
+        String home = System.getProperty("user.home");
```
stephlewyh: `println` statements alternate between methods. Could you consider any way if [SLAP](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/codeQuality.html#slap-hard) could be maintained?

src/main/java/duke/Duke.java
heyjinwei: thank you for the comment, Stephanie. May I know how I could abstract the exception handling away? Should I include them within the method which caused the exception? Because currently I'm throwing the exception from within those methods and catching it to print an error message outside. I think that is what is causing the method to be so long.

src/main/java/duke/Duke.java
heyjinwei: Switched it around to make the happy path prominent. Thank you for the advice

src/main/java/duke/Duke.java
heyjinwei: I have corrected some of it, but similar to the previous comment, I am unsure if I am suppose to handle the exceptions within the methods

src/main/java/duke/Duke.java
stephlewyh: Hi jinwei, you may want to consider some options: (1) put the try catch block in the method where the exception is likely to occur, or (2) throwing those exceptions from the method you think the exception might occur. See checked exceptions: [https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/cppToJava.html#how-to-use-exceptions](https://nus-cs2113-ay2021s2.github.io/website/se-book-adapted/chapters/cppToJava.html#how-to-use-exceptions).

PR #91, opened 2021-02-03
[Zhang Yongzhe] iP

zhangyongzhe20:


src/main/java/Duke.java
```diff
@@ -17,23 +17,40 @@ public static void main(String[] args) {
...
         String userInput;
         String output;
-        List<String> tasks = new ArrayList<>();
+        List<task> tasks = new ArrayList<>();
         do{
             userInput = sc.nextLine();
+
+            int taskIndex = 0;
+            if(userInput.contains("done")){
+                String[] split_ = userInput.split(" ");
```
JoviYeung92: maybe the name split here can change to splitSentence

src/main/java/Duke.java
```diff
@@ -17,23 +17,40 @@ public static void main(String[] args) {
...
+                userInput = split_[0];
+                taskIndex = Integer.valueOf(split_[1]);
+
+            }
             switch (userInput) {
```
JoviYeung92: well indented switch case over here :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,56 @@
...
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+//        String logo = " ____        _        \n"
+//                + "|  _ \\ _   _| | _____ \n"
+//                + "| | | | | | | |/ / _ \\\n"
+//                + "| |_| | |_| |   <  __/\n"
+//                + "|____/ \\__,_|_|\\_\\___|\n";
+        String greeting = "Hello I'm Duke\n"
```
L-Irvin, line 37: You can consider using "final" to indicate that the value is constant.

src/main/java/Duke.java
```diff
@@ -1,10 +1,56 @@
...
+                    int i = 1;
+                    for(task task : tasks){
+                        System.out.println(i +  "." + task.getStatusIcon() + " " + task.description);
+                        i++;
+                    }
+                    break;
+                case "done":
+                    task task_ = tasks.get(taskIndex);
+                    task_.markAsDone();
+                    System.out.println("Nice! I've marked this task as done:\n" +
```
L-Irvin: Great job on writing a multi-line command as it looks neat!

src/main/java/task.java
```diff
@@ -0,0 +1,17 @@
+public class task {
```
L-Irvin: It would be better to name your classes in capital.

src/main/java/Duke.java
```diff
@@ -17,23 +17,40 @@ public static void main(String[] args) {
...
+                    break;
                 default:
-                    tasks.add(userInput);
+                    // create a new task
+                    task newTask = new task(userInput);
+                    // added to list
+                    tasks.add(newTask);
                     System.out.println("added: " + userInput);
             }
         } while(!userInput.toLowerCase().equals("bye"));
```
JoviYeung92: well indented do while loop :)

src/main/java/Duke.java
```diff
@@ -17,23 +17,40 @@ public static void main(String[] args) {
...
+                taskIndex = Integer.valueOf(split_[1]);
+
+            }
             switch (userInput) {
                 case "bye":
                     output = "Bye. Hope to see you again soon!";
                     System.out.println(output);
                     break;
                 case "list":
                     int i = 1;
```
JoviYeung92: since 'i' is the number of entry number maybe you can create a variable name "numTask" . If you are using the standard forloop the it is fine to use 'i'. Standard forloop i mean (int i; i<a; i++)

JoviYeung92 (commented):
1st review :) thank you

src/main/java/Duke.java
```diff
@@ -1,10 +1,88 @@
...
+import duke.Todo;
+import exception.DukeException;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
 public class Duke {
+    static List<Task> tasks = new ArrayList<>();
+    static Boolean notBye = true;
```
stephlewyh: As good coding practice, you could indicate the access modifiers (`public`, `private`, `protected` etc..) for these variables.

src/main/java/Duke.java
```diff
@@ -1,10 +1,88 @@
...
+                + "What can I do for you?";
+        System.out.println(greeting);
+
+        Scanner sc = new Scanner(System.in);
+        String userInput;
+        while(notBye) {
+            userInput = sc.nextLine();
+            try {
+                processUserInput(userInput);
+            }catch (DukeException e){
```
stephlewyh: It would be good to leave space between the brackets `{`, `}`, for layout consistency. There are several more such adjustments that could be made in the `processInput` method.

src/main/java/Duke.java
```diff
@@ -1,10 +1,88 @@
...
+        String description = null;
+        if(userInput.startsWith("todo")){
+            if(userInput.equals("todo")){
+                throw new DukeException("OOPS!!! The description of a todo cannot be empty.");
+            }
+            else {
+                description = userInput.substring(5);
+                tasks.add(new Todo(description));
+            }
+            }
```
stephlewyh: Though not affecting the runtime, this curly bracket indentation could be misleading at first glance.  It seems you meant for it to close the block starting from line 45 instead of line 44.  Consider reviewing this portion of code to increase readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,88 @@
...
+        }
+    }
+
+    /**
+     * process the user input and understand the command
+     * @param userInput: value input by a user
+     */
+    private static void processUserInput(String userInput) throws DukeException {
+        String description = null;
+        if(userInput.startsWith("todo")){
```
stephlewyh: You are handling a few cases based on the user input. Perhaps the nested `if` and several `else if ` could be handled as cases to switch around.

src/main/java/Duke.java
```diff
@@ -1,10 +1,88 @@
...
+            task_.markAsDone();
+            System.out.println("Nice! I've marked this task as done:\n" +
+                    task_.toString());
+        }
+        else if(userInput.equalsIgnoreCase("bye")){
+                String bye = "Bye. Hope to see you again soon!";
+                System.out.println(bye);
+                notBye = false;
+        }
+        else{
```
stephlewyh: This could be a default case for handling unexpected inputs.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,19 @@
+package duke;
+
+public class Task {
+    protected String description;
```
stephlewyh, line 5: This is a well defined and simple to read class. Again, you could ensure space before the `{`.

PR #106, opened 2021-02-05
[Chiam Jia-En] IP

Chiamjiaen:
Add Greet, Echo, Exit.
Add task, list, mark done, todo, event and deadline.


src/main/java/Deadline.java
```diff
@@ -0,0 +1,16 @@
...
+    protected String by;
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.getDescription() + " (by: " + by + ")";
```
rafaelperes, line 26: maybe you should make the use of constants in order to avoid magic strings here...

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
...
+    protected String at;
+
+    public Event(String description, String at) {
+        super(description);
+        this.at = at;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.getDescription() + " (at: " + at + ")";
```
rafaelperes, line 25: same goes here...refer to line 11 Deadline.java comments

src/main/java/Parser.java
```diff
@@ -0,0 +1,160 @@
...
+            System.out.println("☹ OOPS!!! I am unable to create a new directory or file :-(");
+        }
+        try {
+            Scanner saveFileScanner = new Scanner(savedList);
+
+            while (saveFileScanner.hasNext()) {
+                String currentRead = saveFileScanner.nextLine();
+                String[] taskSave = currentRead.trim().split(" \\| ");
+                switch (taskSave[0].trim()) {
+                    case "todo":
```
rafaelperes: maybe change "todo", "deadline","event" to constants (enum?) in order to remove magic strings. In addition, maybe you should make the use of  Single Level of Abstraction Principle (SLAP) concept and for each case statement devise a method to handle the abstractions.

src/main/java/Parser.java
```diff
@@ -0,0 +1,160 @@
...
+                isValidCommand = isCommand(command);
+            } catch (UnrecognizedCommandException e) {
+                System.out.println("☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
+            if (isValidCommand) {
+                switch (command) {
+                    case "list":
+                        list.printList();
+                        break;
+                    case "done":
```
rafaelperes: same goes here... maybe you should improve your code abstraction by having a single method (in charge for the abstraction) for each case statement

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
...
+    public String toString() {
+        return ("[T][" + getStatusIcon() + "] " + description);
+    }
+
+    public String getDescription() {
+        return ("[" + getStatusIcon() + "] " + description);
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "\u2713" : "\u2718"); //return tick or X symbols
```
rafaelperes, line 43: refer to the magic string comment.

rafaelperes (commented):
Good Job! ;)

PR #111, opened 2021-02-10
[Nagi Teja] ip

nagiteja:


src/main/java/AddCommand.java
```diff
@@ -0,0 +1,60 @@
...
+
+    AddCommand(Tasks[] tasks, String input, int counter, int type) {
+        super(tasks);
+        this.input = input;
+        this.counter = counter;
+        this.type = type;
+    }
+
+    public Tasks[] addTask() {
+        if (type == 0) {
```
Mick609: Using numbers 0, 1, and 2 to represent task type of event, deadline, and todo would be a bit confusing. Please consider changing it.

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,60 @@
...
+            int startIndex = input.indexOf("n");
+            int endIndex = input.indexOf("\\at");
+            String preciseInput = input.substring(startIndex+3,endIndex-1).trim();
+            Deadline temp = new Deadline(preciseInput,input.split(" ")[0],input.substring(endIndex+4));
```
Mick609: Please consider using a better name for intermediate variable instead of "temp". Maybe "temporaryDeadline" or "newDeadline". The name should explain a bit more of what the variable is for.

src/main/java/Command.java
```diff
@@ -0,0 +1,14 @@
...
+    Tasks[] tasks;
+    Command(Tasks[] tasks){
+        this.tasks = tasks;
+    }
+
+    public static void displayLine(){
+        System.out.println("____________________________________________________________");
+    }
+
+    public static void addedMessage(){
```
Mick609: The name of a method better is an action. So maybe this method should be called "displayAddedMessage"?

src/main/java/DoneCommand.java
```diff
@@ -0,0 +1,16 @@
...
+        super(tasks);
+        this.n = n;
+    }
+
+    public void markDone() {
+        displayLine();
+        System.out.println("Nice! I've marked this task as done: ");
+        System.out.println("[" +tasks[n-1].displayTaskType()+ "] [\u2718] " + tasks[n-1].getDescription());
+        displayLine();
+        tasks[n-1].setDone();
```
Mick609: This is just a suggestion. The script executes the "setDone" method after displaying the message "I have marked this task as done". The sequence is ok when everything works fine as there are only probably several milliseconds of delay. But when the "setDone" method causes an error, the program would crash after the message "I have marked this task as done". This would be a bit misleading.

src/main/java/Duke.java
```diff
@@ -1,10 +1,92 @@
...
+                    command.printList();
+                } else if (input.split(" ")[0].equals("done")) {
+                    int num = Integer.parseInt(input.split(" ")[1]);
+                    DoneCommand command = new DoneCommand(inputTasks, num);
+                    command.markDone();
+                } else if (input.split(" ")[0].equals("todo")) {
+                    AddCommand command = new AddCommand(inputTasks, input, counter, 1);
+                    inputTasks = command.addTask();
+                    counter++;
+
```
Mick609: There is some unnecessary space between the lines.

Mick609 (commented):
The code is easy to read and understand while there are only some minor problems. Good job.

